PR #6, opened 2019-09-30
Random subpsace method

rguo123:


multiview/construct/rsm.py
```diff
@@ -46,25 +49,26 @@ def random_subspace_method(X, n_features=None, n_views=1):
...
+
     for _ in range(n_views):
         view = np.copy(X)
         features_selected = []
-        
+
         while len(features_selected) != n_features:
-            
+
             feature = random.randint(0, cols - 1)
```
gmischl1: Should there be a parameter to allow the user to choose the seed for this function?

multiview/construct/rsm.py
```diff
@@ -0,0 +1,125 @@
...
+        Number of views to construct.
+
+    Returns
+    -------
+    views : list of array-like matrices
+        List of constructed views (each matrix has shape [n_rows, n_cols]).
+
+    """
+
+    # todo let max_features be float for probability
```
rflperry: Merge fine if this todo is a future todo

multiview/construct/rsm.py
rflperry: I think it sufficient to set seed before calling the function

multiview/construct/rsm.py
rguo123: Ooh good catch. Already implemented just forgot to delete comment.

PR #7, opened 2019-10-03
Added toy data csv files from UCI multiview dataset

gmischl1:


gmischl1:
The name is Multiple Features Data Set and it comes from UCI, so hopefully UCImultifeature is descriptive enough

PR #12, opened 2019-10-17
Multiview KMeans

achang63:
This pull request includes the MultiviewKMeans class as well as its associated base class, utils, and unit tests.


multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,262 @@
...
+    Paramters
+    ---------
+    k : int
+        The number of clusters
+
+    random_state : int (default=None)
+        Determines random number generation for initializing centroids.
+        Can seed the random number generator with an int.
+
+    max_iter : int (default=None)
```
gmischl1: even though the default is technically none, I think this should probably say default = infinity because that's what the code changes it to

multiview/cluster/base_cluster.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @abstractmethod
+    def fit(self, Xs, y=None):
+
+        '''
+        A method to fit clustering parameters to the multiview data.
+        Parameters
+        ----------
+        Xs: a list of numpy arrays
+        '''
```
gmischl1: Add description of y argument

multiview/utils/cluster_utils.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2019 NeuroData (http://neurodata.io)
+#
```
gmischl1: Is this file needed? Can we put these utils into the utils.py file?

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Parameters
+        ----------
+        Xs : list of array_likes
+            - Xs shape: (2,)
+            - Xs[0] shape: (n_samples, n_features_i)
+            This list must be of size 2, corresponding to the two views
+            of the data. The two views can each have a different number
+            of features, but they must have the same number of samples.
+
```
gmischl1: add descriptions of other input arguments to parameters. It looks like they're in the Returns section?

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,262 @@
...
+        initialization.
+
+    _centroids : list of array_likes
+        - _centroids shape: (2,)
+        - _centroids[0] shape: (n_clusters, n_features_i)
+        The cluster centroids for each of the two views. _centroids[0]
+        corresponds to the centroids of view 1 and _cnetroids[1] corresponds
+        to the centroids of view 2.
+
+
```
gmischl1: Maybe not necessary, but could add a reference to a co-em paper if there's a seminal one

rguo123:
Can you change PR title to multiview kmeans?

multiview/cluster/mv_k_means.py
rguo123: This doesn't seem to a class parameter? Looks like it's only called on the function level. I would remove here. Otherwise, I would set max_iter in the constructor and remove the parameter from the functions (this is what sklearn does: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/cluster/_k_means.py so might be preferred)

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,262 @@
...
+            raise ValueError(msg)
+
+        if max_iter is not None:
+            if not (isinstance(max_iter, int) and (max_iter > 0)):
+                msg = 'max_iter must be a positive integer'
+                raise ValueError(msg)
+        else:
+            max_iter = np.inf
+
+        indices = np.random.choice(Xs[1].shape[0], self._k)
```
rguo123: inline comment here specifying that initial centroids are chosen randomly

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,262 @@
...
+
+        while(iter_stall < patience and iter_num < max_iter):
+            iter_num += 1
+            view = (iter_num + 1) % 2
+
+            # Switch partitions, maximization, and expectation
+
+            new_centers = list()
+            for cl in range(self._k):
+                mask = (partitions[(view + 1) % 2] == cl)
```
rguo123: Can you add an inline comment here? Apply to other places where there's a decent amount of steps happening on one line (due to a python/numpy trick).

multiview/utils/cluster_utils.py
rguo123: I think the difference in his utils is that the need for two views is enforced? @achang63 lmk if this isn't true.

What might be better is to remove this file and just have that additional check be part of Kmeans? It's also fine to have your own utils file for later functions but simplify it so that there isn't code repetition between utils/utils.py and this file.

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,262 @@
...
+            function after which the algorithm will terminate.
+
+        max_iter: int, optional (default=None)
+            The maximum number of EM iterations to run before
+            termination.
+
+        '''
+
+        self.fit(Xs, patience, max_iter)
+        partitions = self.predict(Xs)
```
rguo123: rename this to predictions since that's what you are returning?

multiview/utils/cluster_utils.py
gmischl1: If you need 2 views enforced, that can be combined with the new utils, because the CTClassifier commit added arguments to the Check_Xs util to enforce a certain number of views

rguo123:
General comment: test coverage should be a lot higher

multiview/cluster/mv_k_means.py
achang63: I think I'll leave the default as none here since scikit-learn's API does the same for regular kmeans.

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,297 @@
...
+
+        for cl in range(self._k):
+            dist = X - centers[cl]
+            dist = np.linalg.norm(dist, axis=1)
+            distances.append(dist)
+
+        distances = np.vstack(distances)
+        return distances
+
+    def fit(self, Xs, patience=5, max_iter=None):
```
bdpedigo: see my comment in slack about having parameters passed into fit

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,344 @@
...
+        function after which the algorithm will terminate.
+
+    max_iter : int, optional (default=None)
+        The maximum number of EM iterations to run before
+        termination.
+
+    Attributes
+    ----------
+
+    _n_clusters : int
```
bdpedigo: None of your parameters should be listed as attributes here, see SKlearn API

Also unsure why these are private attributes (the leading underscore)

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,344 @@
...
+    [1] Bickel S, Scheffer T (2004) Multi-view clustering. Proceedings of the
+    4th IEEE International Conference on Data Mining, pp. 19–26
+    '''
+
+    def __init__(self, n_clusters=2, random_state=None,
+                 patience=5, max_iter=None):
+
+        super().__init__()
+
+        if not (isinstance(n_clusters, int) and n_clusters > 0):
```
bdpedigo: note that in the sklearn API typechecking is done on fit, not init. In graspy we did it this way also, however, but if you actually want to be super compliant then you should move this stuff

I think the only thing it actually matters for in practice is with some of the pipeline stuff they have in sklearn. Have a look at the link i sent in slack and decide what you all want to do as a group

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,344 @@
...
+            raise AttributeError(msg)
+
+        dist1 = self._compute_distance(Xs[0], self._centroids[0])
+        dist2 = self._compute_distance(Xs[1], self._centroids[1])
+        dist_metric = dist1 + dist2
+        predictions = np.argmin(dist_metric, axis=0).flatten()
+
+        return predictions
+
+    def fit_predict(self, Xs):
```
bdpedigo: Should this go in base class?

achang63: Do you mean the implemented method or the abstract method?

multiview/cluster/mv_k_means.py
achang63: Ok

PR #13, opened 2019-10-17
Created Dataloader function for UCImultifeature dataset and unit tests

gmischl1:


tests/test_dataloader.py
```diff
@@ -0,0 +1,48 @@
...
+
+def test_UCImultifeature_dataloader_badselect2():
+    long_list = [0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+    with pytest.raises(ValueError):
+        data, labels = load_UCImultifeature(select_labeled=long_list)
+
+def test_UCImultifeature_dataloader_badselect3():
+    bad_list = [0,2,4,-2]
+    with pytest.raises(ValueError):
+        data, labels = load_UCImultifeature(select_labeled=bad_list)
```
rflperry: Newline might be required?

gmischl1:
I didn't even realize the notebook was still in this PR. I didn't mean to include that at all. I'll make a new one as an example of how to use this and include it in the examples folder

tests/test_dataloader.py
gmischl1: done

PR #15, opened 2019-10-27
Random gaussian projection

rguo123:
This PR wraps sklearn's random gaussian projection for view construction. This PR also includes several changes to code and directory structure so that view construction is better modularized.


rguo123:
check_n_features is used in random subspace method. I just refactored utils out of random subspace methods since it made more sense.

PR #16, opened 2019-10-28
Multiview MDS Code

akoul1:
First PR for Multiview MDS.


rguo123:
Wrong module. Put this in the embedding module

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+import warnings
+import numpy as np
+from sklearn.metrics import euclidean_distances
+
+
+class MVMDS(BaseEmbed):
+
+    """"
+    Classical Multiview Multidimensional Scaling for jointly reducing
+    the dimensions of multiple views of data. A euclidean distance matrix
```
rguo123: capitalize E in Euclidean

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+
+    """
+
+    def __init__(self, n_components=None):
+
+        super().__init__()
+        self.components = None
+        self.n_components = n_components
+
+    def _cpc(self, x):
```
rguo123: rename x to Xs like GCCA so we have a similar format

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+        it = 15
+
+        for i in np.arange(views):
+            s = s + (n_num[i] * x[i])
+
+        e1, e2 = np.linalg.eigh(s)
+
+        q0 = e2[:, ::-1]
+
+        for i in np.arange(self.n_components):
```
rguo123: add in line comments throughout this function. there's a lot happening and it's hard to read/understand what the code is doing without them

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+                d[:, j] = np.dot(np.dot(q.T, x[j]), q)
+
+            for j in np.arange(it):
+                s2 = np.zeros((p, p))
+
+                for yy in np.arange(views):
+                    d2 = n_num[yy] * np.sum(np.array([n] * views))
+
+                    if d[:, yy] == 0:
+                        s2 = s2 + (d2 * x[yy] / .0001)
```
rguo123: What's happening here with the .0001 constant?

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+            Number of components requested
+
+        Attributes
+        ----------
+
+        components: A k-dimensional projection of shape [n,k]
+
+        """
+
+        if (self.n_components) > len(Xs[0]):
```
rguo123: For the future, all these checks should be put into a private function and then called here. However, for the sake of the time I think it's fine as is. Maybe add an issue for now

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Parameters
+        ----------
+
+        Xs: list of array-likes
+                - Xs shape: (n_views,)
+                - xXs[i] shape: (n_samples, n_features_i)
+                The data to fit to. Each sample will receive its own embedding.
+
+        n_components: int (positive), optional, default = 0
```
rguo123: n_components is not a parameter here anymore

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+        by common Trendafilov implementations based on the following paper:
+
+        https://www.sciencedirect.com/science/article/pii/S016794731000112X
+
+        Parameters
+        ----------
+
+        x: List of matrices, each with number of rows, n
+
+        it: Number of common principal component stepwise iterations
```
rguo123: it is no longer a parameter here. If it is, you should move it to class constructor. Also, name it num_iter so it's more clear

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Returns
+        -------
+
+        Components: Desired Common Principal Components
```
rguo123: Nit: don't capitalize components. Change here and in rest of function

rguo123: From Pedigo's comment on my PR, fit should not be returning anything just setting the embeddings as an attribute (e.g. embeddings_). Then, write a fit_transform function that calls fit and also returns self.embeddings_ (look here for an example https://github.com/NeuroDataDesign/multiview/pull/34)

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+
+    """"
+    Classical Multiview Multidimensional Scaling for jointly reducing
+    the dimensions of multiple views of data. A euclidean distance matrix
+    is created for each view, double centered, and the k largest common
+    eigenvectors are returned based on the algorithm proposed by the
+    following paper:
+
+    https://www.sciencedirect.com/science/article/pii/S016794731000112X
+
```
rguo123: In this top level comment, specify parameters and attributes used. For example, n_components should go here and number of iterations.

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+        n_num = np.array([n] * views)/np.sum(np.array([n] * views))
+
+        Components = np.zeros((p, self.n_components))
+
+        pi = np.eye(p)
+
+        s = np.zeros((p, p))
+
+        it = 15
```
rguo123: No, don't hardcode 'it'. Leave it as a class parameter (see gcca or omnibus as an example on setting class parameters). Note that you will need code to check that it is valid, positive int.

rguo123: You can leave 15 as a default value for 'it'

multiview/embed/mvmds.py
akoul1: It is to prevent divide by 0 error. If views are very dependent on each other, will not run, this way it spits back an NA instead

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        Finds Stepwise Estimation of Common Principal Components as described
+        by common Trendafilov implementations based on the following paper:
+
+        https://www.sciencedirect.com/science/article/pii/S016794731000112X
+
+        Parameters
+        ----------
+
+        x: List of matrices, each with number of rows, n
```
gmischl1: after you rename to Xs, make this the same docstring format as here so all our code is consistent https://github.com/NeuroDataDesign/multiview/pull/45/files

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+    def fit(self, Xs):
+
+        """
+        Calculates a dimensional reduction based on minimizing Euclidean
+        Distances of multiple views
+
+        Parameters
+        ----------
+
+        Xs: list of array-likes
```
gmischl1: Change slightly to be same format as https://github.com/NeuroDataDesign/multiview/pull/45/files

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+        x: List of matrices, each with number of rows, n
+
+        it: Number of common principal component stepwise iterations
+
+        Returns
+        -------
+
+        Components: Desired Common Principal Components
+
+        """
```
gmischl1: use our utils.check_Xs for basic input validation

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+
+        """"
+        Embeds data matrix(s) using fitted projection matrices
+
+        Returns back the original Xs. Not needed for implementation
+
+        Parameters
+        ----------
+
+        Xs: list of array-likes
```
gmischl1: change slightly to match https://github.com/NeuroDataDesign/multiview/pull/45/files

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,217 @@
...
+    def fit_transform(self, Xs):
+
+        """"
+        Embeds data matrix(s) using fitted projection matrices
+
+        Parameters
+        ----------
+
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
```
gmischl1: change slightly to match https://github.com/NeuroDataDesign/multiview/pull/45/files

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __init__(self, n_components=None, num_iter=15):
+
+        super().__init__()
+        self.components = None
+        self.n_components = n_components
+        self.num_iter = num_iter
+        
+
+    def _cpc(self, Xs):
```
bdpedigo: could have a better name for this

multiview/embed/mvmds.py
bdpedigo: or better yet maybe break this up into smaller functions if possible?

bdpedigo: with descriptive names

multiview/embed/mvmds.py
bdpedigo: + 1

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,240 @@
...
+            view = euclidean_distances(Xs[i])
+            view_squared = np.power(np.array(view), 2)
+
+            J = np.eye(len(view)) - (1/len(view))*np.ones(view.shape)
+            B = -(1/2) * np.matmul(np.matmul(J, view_squared), J)
+            mat[i] = B
+
+        self.components = self._cpc(mat)
+
+    def transform(self, Xs):
```
bdpedigo: is this for API compliance somehow?

bdpedigo: I don't know how I feel about this when the transform doesn't make sense for out of sample data, people might get confused and try to use this.

bdpedigo:
I feel like you need more proof for this, I didn't get a good sense that the thing was working. Did they have no simulations in the original paper, I forget?

In any case, I'd have a simulated data example, maybe something along the lines of several clusters in 2 different views, clusters are a bit noisy/overlapping so classification would be hard in 1 view, but should be able to find a setting where if you MVMDS both views, and then cluster, get a fairly good result. Compare this to clustering either view (after regular MDS?) separately, or the naive multiview way, just concatenating features and then embedding and then clustering. 

Based on the above I'll have a better sense of when/why/how this thing is helpful. Also, should be able to come up with settings where MVMDS _doesn't_ help, we want to know what those are too

multiview/embed/mvmds.py
akoul1: would commonpc be better?

multiview/embed/mvmds.py
akoul1: Will add as an issue

multiview/embed/mvmds.py
akoul1: Had talked to Ronan about this. I think we decided to do this to keep with the general embedding format. Couldn't split up the function as such. I can keep this as an issue and try to revisit it. I think this relates to making the larger _commonpcs into smaller pieces. This may be a bit of a process because I wrote that function straight from the pseudocode outlined in the paper.

multiview/embed/mvmds.py
akoul1: Will keep splitting it up as an issue. I wrote this function straight from the pseudocode trendafilov gave in his paper. Can think about how to go about splitting it up

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,239 @@
...
+        traditional, single-view Multidimensional Scaling.
+
+        Parameters
+        ----------
+
+        Xs: list of array-likes or numpy.ndarray
+                - Xs length: n_views
+                - Xs[i] shape: (n_samples, n_features_i)
+
+        Attributes
```
rguo123: Attributes should be described in the overall class comment.

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,239 @@
...
+        ----------
+
+        components: A k-dimensional projection of shape [n,k]
```
rguo123: Need to specify what n and k are here

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,239 @@
...
+        """"
+        Embeds data matrix(s) using fitted projection matrices
+
+        Parameters
+        ----------
+
+        Xs: list of array-likes or numpy.ndarray
+            - Xs length: n_views
+            - Xs[i] shape: (n_samples, n_features_i)
+            The data to embed based on the prior fit function.
```
rguo123: Dont have the word 'prior' here

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Returns
+        -------
+        components: Components of the dimensionally reduced Xs
```
rguo123: this phrasing sounds a little funny. Also the format is the type of the return and then below it you have a description.

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Xs: list of array-likes or numpy.ndarray
+            - Xs length: n_views
+            - Xs[i] shape: (n_samples, n_features_i)
+            The data to embed based on the prior fit function.
+
+        Returns
+        -------
+
+        Xs = Same Xs as inputted
```
rguo123: You can remove the transform() function. There is no real reason for keeping this code here so far as i can tell.

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,239 @@
...
+        ----------
+
+        Xs: List of array-likes or numpy.ndarray
+            - Xs length: n_views
+            - Xs[i] shape: (n_samples, n_features_i)
+
+        Returns
+        -------
+
+        components: The first n_component primary principal components
```
rguo123: Format the return correctly. Aka first line should be the return type and below it is the description.

rguo123: Specifying the shape is also important here

multiview/embed/mvmds.py
rguo123: Just in case the docs order the functions differently or we change the order

multiview/embed/mvmds.py
rguo123: Didn't see this being addressed?

multiview/embed/mvmds.py
```diff
@@ -0,0 +1,216 @@
...
+        be returned from the algorithm. This value must be greater than
+        0 and less than the number of samples within each view.
+
+    num_iter: int (positive), default=15
+        Number of iterations stepwise estimation goes through. Detailed
+        in Trendafilov paper.
+
+    Attributes
+    ----------
+    components: A projection of shape [n_samples, n_components]
```
rguo123: Format this correctly (components: return type and then description below). Having the shape here is good

bdpedigo:
> @armankoul I think the new notebook is definitely an improvement, and, maybe a clearer tutorial than the tutorial, IMO, but I leave that up to you guys!
> 
> A few comments
> 
> * i never trust single runs of k-means, there is a ton of variance from the initializations, and the difference between PCA and MVMDS here isn't big enough that I think this should be ignored (long term, i'm just saying look into it at some point, doesn't have to be for this PR).
> * you should say what the clusters you are plotting are (true vs predicted). At first glance I thought these were predicted. showing two plots side by side, one true, one predicted, i think is informative,

Re-address point #1 ?

PR #17, opened 2019-11-04
Kcca

theohj826:
PR for kcca. Not sure if I have to make sure my branch has to be up to date with master before I PR. Please let me know if I did anything else wrong (organizing files, etc.)


rguo123:
Make sure to put kcca tutorial with docs.

multiview/embed/kcca.py
```diff
@@ -0,0 +1,485 @@
...
+Python module for regularized kernel canonical correlation analysis.
+Code adopted from UC Berkeley, Gallant lab
+(https://github.com/gallantlab/pyrcca)
+"""
+
+import numpy as np
+from scipy.linalg import eigh
+
+
+class KCCA(object):
```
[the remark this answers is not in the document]
bdpedigo: + 1

multiview/embed/kcca.py
```diff
@@ -0,0 +1,488 @@
...
+        kernelcca=True,
+        ktype=None,
+        verbose=False,
+        cutoff=1e-15,
+        gausigma=1.0,
+        degree=2,
+    ):
+        self.reg = reg
+        self.numCC = numCC
+        self.kernelcca = kernelcca
```
bdpedigo: why would we want this to be an option?

multiview/embed/kcca.py
```diff
@@ -0,0 +1,488 @@
...
+        self.ktype = ktype
+        self.cutoff = cutoff
+        self.gausigma = gausigma
+        self.degree = degree
+        if self.kernelcca and self.ktype is None:
```
bdpedigo: should just make a default kwarg

multiview/embed/kcca2.py
```diff
@@ -0,0 +1,312 @@
...
+        calculated using SVD. The number of singular vectors kept is determined
+        by either the percent variance explained, a given rank threshold, or a
+        given number of components. The singular vectors kept are concatenated
+        and SVD of that is taken and used to calculated projections for each
+        view.
+
+        Parameters
+        ----------
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
```
gmischl1: Change this docstring slightly to match https://github.com/NeuroDataDesign/multiview/pull/45/files

multiview/embed/kcca2.py
```diff
@@ -0,0 +1,312 @@
...
+        return self
+
+    def transform(self, Xs, view_idx=None):
+        """
+        Embeds data matrix(s) using the fitted projection matrices
+
+        Parameters
+        ----------
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
```
gmischl1: same as above comment

multiview/embed/kcca2.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        Fit to data, then transform it.
+
+        Fits transformer to Xs optional parameters fit_params and returns a
+        transformed version of the Xs.
+
+        Parameters
+        ----------
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
```
gmischl1: same as above comments

multiview/embed/test.py
```diff
@@ -0,0 +1,59 @@
+import numpy as np
+
+
```
gmischl1: What is this file for?

gmischl1:
Why are there 3 kcca files? kcca.py, kcca2.py, and kcca3.py?

bdpedigo:
Linear Kernel CCA Tutorial/Proof
- `sys.path.append("../../..")` should not be necessary, if it is, i think something else is wrong. not going to harp on this for this PR but you guys should figure out how to not use this
- plot the data before training as well, its not very informative to only see output
- don't really need ticks/ticklables here IMO
- the plots of component 1 vs component2 don't really do anything for me, but maybe i'm just not getting it?
- cell output 17 how are the two plots different, the labels are the same 
- after looking at the proof notebook it seems like there were many things we talked about that you did not address: 
     - What about running multiple experiments for each of the 2 simulation experiments in the kcca paper, and plotting all of your results relative to the paper's, and maybe do the same for that other implementation? 
     - You need more simulation examples, showing that your thing works sometimes, and doesn't work sometimes. See Gavin's proof notebook. This will help develop an intuition for the geometries where it works/doesnt work and that will be really informative 
     - I would compare against regular CCA (sklearn) as well 
     - When would I use linear-KCCA instead of CCA? That's what this should help answer.

 - I would have done separate proof/tutorial but I won't be strict about that now. but eventually you guys need a short tutorial only notebook for the docs

multiview/embed/test.py
theohj826: those were just testing things. i removed them now

multiview/embed/kcca.py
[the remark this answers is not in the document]
theohj826: Should i be extending the base class? I wasnt aware. Will work on this.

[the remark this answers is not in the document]
theohj826: completed this

multiview/embed/kcca.py
theohj826: completed. changed it before you noted it.

multiview/embed/kcca.py
```diff
@@ -0,0 +1,485 @@
...
+    preds_
+        Predictions on the validation dataset (List)
+    """
+    iws = [np.linalg.pinv(w.T, rcond=cutoff) for w in ws_]
+    ccomp = _listdot([d.T for d in vdata], ws_)
+    ccomp = np.array(ccomp)
+    preds_ = []
+    corrs_ = []
+
+    for dnum in range(len(vdata)):
```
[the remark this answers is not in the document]
theohj826: im sure this could be faster, but this is just for validation. not a core component of the kcca class that will be used.

multiview/embed/kcca.py
```diff
@@ -0,0 +1,485 @@
...
+        ktype=None,
+        verbose=False,
+        cutoff=1e-15,
+        gausigma=1.0,
+        degree=2,
+    ):
+        self.reg = reg
+        self.numCC = numCC
+        self.kernelcca = kernelcca
+        self.ktype = ktype
```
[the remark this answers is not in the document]
theohj826: fixed and wrote unit tests

bdpedigo:
> - plot the data before training as well, its not very informative to only see output

bdpedigo:
@theohj826 did i miss this?

bdpedigo:
my takeaway from the notebook is that it's the same as CCA, is that wrong?

PR #33, opened 2019-11-22
Corrected PLS projection weights

rflperry:
Projection weights were incorrectly calculated past the first component. See [link](https://learnche.org/pid/latent-variable-modelling/projection-to-latent-structures/how-the-pls-model-is-calculated) for an understanding of why.


multiview/embed/pls.py
```diff
@@ -49,7 +50,10 @@ def partial_least_squares_embedding(
     """
     pls = PLSRegression(n_components=n_components)
     pls.fit(X, Y)
+    W = pls.x_weights_
+    P = pls.x_loadings_
+    R = W @ np.linalg.pinv(P.T @ W)
```
rguo123: I'm going to go ahead and approve this but can you add some comments on what x_weights and x_loadings are and what the numpy line is doing? Hard to decipher from looking at the code

PR #35, opened 2019-11-22
Simulated performance of cotraining classification

gmischl1:
This is the second part of a PR for cotraining classification. The first part is https://github.com/NeuroDataDesign/multiview/pull/10. This PR includes a notebook detailing the performance of cotraining classification on simulated data. It is meant to support the previously merged PR https://github.com/NeuroDataDesign/multiview/pull/10 and solve Issue https://github.com/NeuroDataDesign/multiview/issues/37.

I also changed the name of the cotraining_classification basic example tutorial noteobook so that it is more clear what it is for.


bdpedigo:
- You should add some more commentary above cells of code, for example, describe in words what you are doing when you generate the first simulated dataset. Explain why you are randomly removing labels. Explain what the single view algorithm is that you are comparing to.
 - Is it fair to compare to a classifier that basically saw half of the data, aren't the two views identical settings in the first simulation? What would happen if you did the naive thing and just concatenated the dimensions of view 1 and view 2 and then trained in 1-view mode, or did PCA first and then trained, etc.
 - What's the logic behind `view2_penalty`?
 - At a glance I don't understand by what you mean by "one randomization", just a train test split? or which algorithm is labeling class 0/1 in the figures for the first sim. 
 - In all of your settings, cotraining wins right? Don't think that should be the case, I'd think more about settings where it shouldn't help (just adds variance) or you're not giving the other algorithm a fair chance (see above maybe?). 
 - Code is really verbose since it looks like you copied/pasted stuff a bunch, needs to be a lot cleaner for a tutorial/demo
 - You don't show the data for the real data part so it's not that illustrative. 
 - I'm not really sure what the corrupted labels is telling us, other than that you aren't cheating when computing accuracy, and I think this notebook is getting too long. I know we wanted negative controls, but I think the overlapping classes sim. you did is more informative. Also, how about something where the two views are truly identical, would that be interesting?
 - I wouldn't have printed output that isn't labeled in the printout and ideally, described in text above the cell.

gmischl1:
I have tried to address many of comments with an updated notebook.

1. I added descriptions/commentary above each simulation to describe the experimental setup.
2. For each simulation, I added a test of a classifier trained on a concatenation of the two views.
3. I specify "view2_penalty" in order to use classifiers with different parameters for each view. A key aspect of cotraining is maintaining independence of the classifications in the two views and one way that the paper does this is by using distinct classifiers for each (like using different distance metrics if using 2 KNN's as the paper does)
4. I attempted to describe better what "one randomization" is. It is supposed to be a single set of N samples from the 2 class distributions which are trained and tested on. The simulation draws N samples 20 times, and the results are the average of the results from the 20 randomizations.
5. I updated the "overlapping" simulation so that now they are perfectly overlapping. Before, they were slightly offset, which allowed cotraining to win by a tiny bit. Now it is a little worse than one of the views and basically performs with chance. Also, now in the new redundant view test, cotraining is not the best.
6. I created a couple helper functions for data creation and plotting that each simulation uses so they are a little less verbose.
7. I kept in the corrupted labels test for now, but we can talk about that. I added a test where the views hold identical data, and cotraining does worse than single view.
8. I removed printouts.

bdpedigo:
- I saw a new comment that hinted at this but I think said it less explicitly in the notebook, might be a good point to make more clearly there
> A key aspect of cotraining is maintaining independence of the classifications in the two views and one way that the paper does this is by using distinct classifiers for each (like using different distance metrics if using 2 KNN's as the paper does)

- The title of one of your plots says "CTC performs at chance" but I'd just point out that "All algorithms perform near chance", it's not that you're really doing worse! 
- Still feels like you have a lot of redundant code that makes the notebook way longer than it needs to be.
- Can remove ticks and ticklabels when you plot the data, I think 
- I got really used to MV being red throughout the notebook so I was confused at first by the final plot, where it switches to blue. Sounds minor, but in general it's worth considering using a consistent palette 

Overall, nice work. I really like the new controls you have added, well done!

rguo123:
Is this notebook included in Cotraining 2.0 now?

gmischl1:
> Is this notebook included in Cotraining 2.0 now?

Yes it is.

rflperry:
There seem to be conflicting files? Edit so we can resolve this PR or is this PR out of date?

PR #39, opened 2019-12-02
implement splitAE with fully connected encoder / decoder networks

cameronfr:
Implements SplitAE as described in this [DCCAE paper](SplitAutoencoderSKLearnAPI.py) with basic fully connected networks (i.e. fixed number of hidden nodes per hidden layer, layers are all linear and not e.g. convolutional).

Here's my notebook with a tutorial and validation: [link](https://github.com/NeuroDataDesign/multiview/blob/cameronfranz/SplitAE%20Example%20and%20Validation.ipynb)  |  [alt link](https://nbviewer.jupyter.org/github/NeuroDataDesign/multiview/blob/cameronfranz/SplitAE%20Example%20and%20Validation.ipynb)


rflperry:
Couple questions/comments
1) Is it possible to make a quick tutorial notebook with the datasets in our package? Just wondering
2) Ideally the test coverage could be higher but is that just an issue of not wanting to train a network while running tests?
3) Is it possible to make optional package dependencies? Like if people don't have torch they just get some sort of warning telling them they need to install. I feel like requiring torch comes with a lot of baggage.

cameronfr:
We could definitely make notebooks. UCIMultifeature is pretty similar to MNIST but other datasets when we add more would be interesting.

The tests currently do make a network and make sure it overfits. Actually the low test coverage is just because  a dozen lines of graphing / print output code is not running in the tests (if they are enabled, pytest will wait for the user to close a popup graph window)

I saw a way to do optional dependencies, but I think requiring pytorch might actually be fine now (pip install torch --no-cache only took 30s total to install including download). But it might be prohibitive to users with slow download speed.

multiview/embed/splitae.py
```diff
@@ -0,0 +1,206 @@
...
+        self.batchSize = batchSize
+        self.learningRate = learningRate
+
+    # Xs is not a tensor but instead a list with two arrays of shape [n, f_i]
+    def fit(self, Xs, validationXs=None, printInfo=True):
+        """
+        Given two views, create and train the autoencoder.
+        Parameters
+        ----------
+        Xs: a list with two arrays. Each array has `n` rows (samples) and some
```
gmischl1: Maybe make this consistent wording with how @rflperry started doing it, just so our package is as uniform as possible. https://github.com/NeuroDataDesign/multiview/pull/45/files

multiview/embed/splitae.py
```diff
@@ -0,0 +1,206 @@
...
+        number of columns (features). The first array is View1 and the second
+        array is View2.
+        """
+
+        assert len(Xs) == 2, "this SplitAE implementation deals with two views"
+        assert Xs[0].shape[0] == Xs[1].shape[0], """must have each view for
+            each sample"""
```
gmischl1: These 2 input validation steps can be done with our utils.check_Xs method

multiview/embed/splitae.py
```diff
@@ -0,0 +1,206 @@
...
+                                            view2Batch.to(device))
+            totalError = view1Error + view2Error
+        return totalError.item()
+
+    def transform(self, Xs):
+        """
+        Transform the given view with the trained autoencoder.
+        Parameters
+        ----------
+        Xs: a list with one array representing the View1 view of some data.
```
gmischl1: Same as above

multiview/embed/splitae.py
```diff
@@ -0,0 +1,206 @@
...
+        return (embedding.cpu().numpy(), view1Reconstruction.cpu().numpy(),
+                view2Prediction.cpu().numpy())
+
+    def fit_transform(self, Xs):
+        """
+        `fit(Xs)` and then `transform(Xs[:1])`. Note that this method will be
+        embedding data that the autoencoder was trained on.
+        Parameters:
+        ----------
+        Xs: see `fit(...)` Xs parameters
```
gmischl1: same as above

cameronfr: Have changed the other docstring wordings to be more consistent! Would rather keep this one as is to prevent duplicating text

bdpedigo:
@cameronfr 

I see that you tried to recreate that figure up to some randomness - in the right direction, but didn't we talk about getting/running their code? 

In any case, I would like to see some simulated experiments as I have been suggesting to everyone in the group. At least one where SplitAE does better than other methods/learning on single views/learning on concatenated views, and at least one where it does worse. 

Also, I thought you were going for DCCAE instead of SplitAE, am I remembering correctly?

bdpedigo:
@cameronfr 

I see that you tried to recreate that figure up to some randomness - in the right direction, but didn't we talk about getting/running their code? 

In any case, I would like to see some simulated experiments as I have been suggesting to everyone in the group. At least one where SplitAE does better than other methods/learning on single views/learning on concatenated views, and at least one where it does worse. 

Also, I thought you were going for DCCAE instead of SplitAE, am I remembering correctly?

bdpedigo:
I'd turn off the progress bars and printouts during training, makes notebook hard to read

multiview/embed/splitae.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def paramCount(self):
+        return np.sum([np.prod(s.shape) for s in self.parameters()])
+
+
+class SplitAE(BaseEmbed):
+    """
+    Implements an autoencoder that creates an embedding of a view View1 and
+    from that embedding reconstructs View1 and another view View2.
+    Parameters
```
bdpedigo: need spacing between these things to get to render properly, look at some of the graspy docs for examples

multiview/embed/splitae.py
```diff
@@ -0,0 +1,209 @@
...
+    hiddenSize: number of nodes in the hidden layers
+    numHiddenLayers: number of hidden layers in each encoder or
+        decoder net
+    embedSize: size of the bottleneck vector in the autoencoder
+    trainingEpochs: how many times the network trains on the full
+        dataset
+    learningRate: learning rate of the Adam optimizer
+    Attributes
+    ----------
+    view1Encoder: the View1 embedding network as a PyTorch module
```
bdpedigo: if these are computed attributes then need trailing underscore

multiview/embed/splitae.py
```diff
@@ -0,0 +1,209 @@
...
+                 trainingEpochs=10, batchSize=16, learningRate=0.001):
+        self.hiddenSize = hiddenSize
+        self.embedSize = embedSize
+        self.numHiddenLayers = numHiddenLayers
+        self.trainingEpochs = trainingEpochs
+        self.batchSize = batchSize
+        self.learningRate = learningRate
+
+    # Xs is not a tensor but instead a list with two arrays of shape [n, f_i]
+    def fit(self, Xs, validationXs=None, printInfo=True):
```
bdpedigo: what are the other parameters, they are not documented? I'd remove them unless you have a very very good reason to keep. Print info can definitely be in the init

multiview/embed/splitae.py
```diff
@@ -0,0 +1,209 @@
...
+
+class SplitAE(BaseEmbed):
+    """
+    Implements an autoencoder that creates an embedding of a view View1 and
+    from that embedding reconstructs View1 and another view View2.
+    Parameters
+    ----------
+    hiddenSize: number of nodes in the hidden layers
+    numHiddenLayers: number of hidden layers in each encoder or
+        decoder net
```
bdpedigo: none of these kwargs follow python naming conventions, need snake_case

tests/embed/test_splitae.py
```diff
@@ -0,0 +1,17 @@
+from multiview.embed.splitae import SplitAE
+import numpy as np
+
+def test_splitae_overfit():
```
bdpedigo: this is not enough testing at all. need input validation at a bare minimum. I also dont understand what looking at the standard deviation here is telling you about the success of the method

bdpedigo: so you also need some kind of better test that your thing is generating a good embedding, not just that it is overfitting...

bdpedigo:
@cameronfr  main two comments on code itself
 - Variable naming does not match python naming conventions, most variables and functions need to be snake_case like in the rest of the package
 - need way more testing, see my comments, and also look at some of the other PRs from your group to see what people's testing suites look like.

tests/embed/test_splitae.py
cameronfr: stdev check is of (actual-predicted) to assure overfitting while allowing for some outliers (e.g. `np.all(np.abs(actual-predicted) < threshold)` would fail sometimes)

cameronfr: Overfit check is usually the first thing that's done when making sure a network is working but maybe I can do something where I set the torch seed and look for certain results

bdpedigo: my point is, why does this metric show us that the method is working? I don't think overfitting is a good test

bdpedigo: * i should say, it isn't sufficient as the only test

cameronfr: I agree that it's not sufficient, I think a better test is looking at the Noisy MNIST embeddings as in the notebook, maybe I can squeeze that into this test suite somehow.

cameronfr: FTR, our other testing suites (at least. [test_omnibus](https://github.com/NeuroDataDesign/multiview/blob/8000bd061c8506daa8d24662ac44ec6dcb85f8b9/tests/test_omnibus.py )) are not testing that a good embedding is found. (e.g.  our Omnibus embedding tests are only checking that the number of views input = the number of views output). Despite it seeming like there is a lot of testing there.

cameronfr:
> @cameronfr
> 
> I see that you tried to recreate that figure up to some randomness - in the right direction, but didn't we talk about getting/running their code?
> 
> In any case, I would like to see some simulated experiments as I have been suggesting to everyone in the group. At least one where SplitAE does better than other methods/learning on single views/learning on concatenated views, and at least one where it does worse.
> 
> Also, I thought you were going for DCCAE instead of SplitAE, am I remembering correctly?

No, going for SplitAE has been the plan for a while. As I felt I was tuning hyperparameters to try and get DCCAE to work. For example, their DCCAE code does not have any T-SNE related code in it, and I found that by tweaking the number of CCA components fed to T-SNE I could get closer to their DCCAE figure, but tweaking numbers like that wasn't fun.

I am sure my SplitAE works correctly (i.e. it is predicted view1, reconstructing view2, and doing something at least somewhat useful). I've added a [notebook](https://github.com/NeuroDataDesign/multiview/blob/cameronfranz/splitae_controls.ipynb) with testing on experimental data (not sure if we are adding these notebooks to our docs folder).

Also, I added another test to the suite to do something more comprehensive than overfitting. 

Also, does anyone know how I can fix the Travis CI saying `Expected — Waiting for status to be reported ` ?

multiview/embed/splitae.py
```diff
@@ -0,0 +1,226 @@
...
+import numpy as np
+import itertools
+import tqdm
+
+from .base import BaseEmbed
+from ..utils.utils import check_Xs
+
+
+class FullyConnectedNet(torch.nn.Module):
+    def __init__(self, inputSize, hiddenSize, numHiddenLayers, embeddingSize):
```
gmischl1: Seems like this class and its functions should have docstrings

multiview/embed/splitae.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class SplitAE(BaseEmbed):
+    """
+    Implements an autoencoder that creates an embedding of a view View1 and
+    from that embedding reconstructs View1 and another view View2.
+
+    Parameters
+    ----------
+    hiddenSize: number of nodes in the hidden layers
```
gmischl1: as Pedigo said below, all variables need to be snake case

multiview/embed/splitae.py
```diff
@@ -0,0 +1,226 @@
...
+    numHiddenLayers: number of hidden layers in each encoder or
```
gmischl1: Also, for parameters and Returns stuf should be formatted as 

var_name : type
        description of parameter or return object

tests/embed/test_splitae.py
```diff
@@ -0,0 +1,62 @@
...
+                for i in range(nClusters):
+                    predictionPermuted[indexes[i]] = permutation[i]
+                accuracy = np.sum(predictionPermuted == target) / len(target)
+                maxAccuracy = accuracy if accuracy > maxAccuracy else maxAccuracy
+            return maxAccuracy
+
+        accuracy = clusterAccuracy(kmeans.labels_, labels[trainSplit:], nClusters=3)
+        accuracies.append(accuracy)
+    # distribution of accuracies sometimes has outliers around 0.65, so be conservative
+    assert np.mean(accuracies) > 0.7
```
gmischl1: Rather than doing this, can you set the seed for the make_blobs function so that you can know exactly how well the algorithm will perform, then assert that the accuracies are within a numerical tolerance of what you expect? Because if in the future you change something else in the algorithm that somehow messes up its ability to predict as well, but it still passes 0.7, it'll pass this test even though its predictive ability changed. Let me know what you think

multiview/embed/splitae.py
cameronfr: It's an internal class so I guess it just needs an underscore

multiview/embed/splitae.py
cameronfr: As mentioned in Slack I'm going to put the in an issue since it's just a matter of style consistency and it's all my code not just parameters

tests/embed/test_splitae.py
cameronfr: Hmmm, I see what you're saying -- that maybe setting the make blobs seed will allow me to set a stricter bound like 0.9 or so. I think the variability in accuracy is due both the variability in make blobs as well as variability in convergence of the model (weights are randomly initialized). But good idea I'll do a test and see if this is the case

cameronfr:
> Added a couple comments about variable names and parameter docstrings, as well as a test.
> 
> Also, in your notebook, you rotate the view 2 images. However, in the paper they don't seem to rotate the view 2 images, right? Why did you rotate them? It seems if you didn't rotate them you might get cleaner looking results as well, since they might not have the hazy ring.
> 
> Also, sorry, but I don't know why Travis CI says that. Is there any way a test could have gotten caught in a loop?

I checked their code and you're right about them not rotating the view2 images as well. I agree that the hazy rings (whos cause was interesting to figure out) will probably disappear. Let me re run the notebook.

[Ronan's PR](https://github.com/NeuroDataDesign/multiview/pull/47/files) is having the same problem with Travis CI and as far as I can tell he's just changing docstrings. Whoever has owner access should try https://github.com/travis-ci/travis-ci/issues/10204 .

gmischl1:
make sure you also change the description of the dataset in the tutorial notebook ("that's similarly rotated" should be removed)

tests/embed/test_splitae.py
cameronfr: So it looks like if I add the seed I can increase the bound a little bit to around 0.8  . But I think the added benefit of testing on different random make_blobs cases (some of which might expose a new weakness if the code is changed detrimentally) outweighs that increase since it's pretty small.

multiview/embed/splitae.py
rguo123: No, provide documentation still so people who are reading or using your code will know what's happening.

multiview/embed/splitae.py
```diff
@@ -0,0 +1,241 @@
...
+        self.layers = torch.nn.ModuleList()
+        if numHiddenLayers == 0:
+            self.layers.append(torch.nn.Linear(inputSize, embeddingSize))
+        else:
+            self.layers.append(torch.nn.Linear(inputSize, hiddenSize))
+            for i in range(numHiddenLayers-1):
+                self.layers.append(torch.nn.Linear(hiddenSize, hiddenSize))
+            self.layers.append(torch.nn.Linear(hiddenSize, embeddingSize))
+
+    def forward(self, x):
```
rguo123: Same for all these functions: provide docs. They can be much less rigorous than public code.

multiview/embed/splitae.py
```diff
@@ -0,0 +1,241 @@
...
+                      .format(epoch, np.mean(batchErrors)))
+            epochTrainErrors.append(np.mean(batchErrors))
+            if validationXs is not None:
+                testError = self._testError(validationXs)
+                if self.printInfo:
+                    print("Average test  error during epoch {} was {}\n"
+                          .format(epoch, testError))
+                epochTestErrors.append(testError)
+
+        if self.printGraph:
```
rguo123: The code here should probably be moved into a separate function for validation testing. Not too important right now though

rguo123:
Code all looks good. Just two things: provide docs for private functions too since people may need to look at this code l8er on and it might get confusing if they dont know what's happening. More in line comments would be nice too but i found it pretty intuitive already. 

At the bare minimum, make all public parameters/attributes that users may set or use to snake case.

multiview/embed/splitae.py
cameronfr: Pytest will pop up a graph so no can do. We can make a test to compare the pixels of the produced graph to a stored reference graph lol

rguo123: srry should have been more clear. Meant that all the print infos should be pulled out as separate helper functions users can call specifically for printing results. Not important right now just thought it'd be better modularized that way.

rguo123: going to go ahead and resolve this for now

multiview/embed/splitae.py
```diff
@@ -0,0 +1,257 @@
...
+            plt.plot(epochTrainErrors, label="train error")
+            if validation_Xs is not None:
+                plt.plot(epochTestErrors, label="test error")
+            plt.title("Errors during training")
+            plt.xlabel("Epoch")
+            plt.ylabel("Error")
+            plt.legend()
+            plt.show()
+
+    def _testError(self, Xs):
```
rguo123: need documentation here

multiview/embed/splitae.py
```diff
@@ -0,0 +1,257 @@
...
+import itertools
+import tqdm
+
+from .base import BaseEmbed
+from ..utils.utils import check_Xs
+
+
+class _FullyConnectedNet(torch.nn.Module):
+    """
+    General torch module for a fully connected neural network with
```
rguo123: Since there are a lot of class parameters here can you just list them out numpy style below? a little hard to read otherwise

multiview/embed/splitae.py
cameronfr: sure, should be pretty self explanatory though

multiview/embed/splitae.py
cameronfr: True

bdpedigo:
add .ipynb file name to simulated data notebook

multiview/embed/splitae.py
```diff
@@ -0,0 +1,259 @@
...
+        self.hiddenSize = hidden_size
+        self.embedSize = embed_size
+        self.numHiddenLayers = num_hidden_layers
+        self.trainingEpochs = training_epochs
+        self.batchSize = batch_size
+        self.learningRate = learning_rate
+        self.printInfo = print_info
+        self.printGraph = print_graph
+
+    def fit(self, Xs, validation_Xs=None):
```
bdpedigo: fit should usually return instance of self

tests/embed/test_splitae.py
```diff
@@ -0,0 +1,62 @@
...
+import sklearn.datasets
+from itertools import permutations
+import torch
+
+# These are blobs that PCA can reduce to nice clusters easily, so splitAE should be able to aswell
+# cluster_std is high s.t. any 2 features alone do not make nice seperable blobs
+def test_splitae_blobs():
+    accuracies = []
+    for i in range(10):
+        view1, labels = sklearn.datasets.make_blobs(n_samples=1000, n_features=20, centers=None, cluster_std=10.0, center_box=(-10.0, 10.0), shuffle=True, random_state=None)
```
bdpedigo: line lengths way too long, i don't know what the standard for this package is

tests/embed/test_splitae.py
bdpedigo: i don't know about changing the bound, however, random seeds for tests like this are pretty essential, otherwise you could be chasing failed builds because of nothing other than poor performance on something random

tests/embed/test_splitae.py
```diff
@@ -0,0 +1,62 @@
+from multiview.embed.splitae import SplitAE
+import numpy as np
+import sklearn.cluster
+import sklearn.datasets
+from itertools import permutations
+import torch
+
```
bdpedigo: should have tests for parameter and argument checking

multiview/embed/splitae.py
```diff
@@ -0,0 +1,259 @@
...
+            number of columns  (features) as the View1 presented
+            in the `fit(...)` step.
+             - Xs length: 1
+             - Xs[0] shape: (n_samples, n_features_0)
+
+        Returns
+        ----------
+        embedding: np.ndarray of shape (n_samples, embeddingSize)
+            the embedding of the View1 data
+        view1Reconstruction: np.ndarray of shape (n_samples, n_features_0)
```
bdpedigo: snake case

multiview/embed/splitae.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Xs = check_Xs(Xs, multiview=True, enforce_views=2)
+        assert Xs[0].shape[0] >= self.batchSize, """batch size must be <= to
+            number of samples"""
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+
+        view1 = torch.FloatTensor(Xs[0])
+        view2 = torch.FloatTensor(Xs[1])
+
+        self.view1Encoder_ = _FullyConnectedNet(view1.shape[1],
```
bdpedigo: snake case

multiview/embed/splitae.py
```diff
@@ -0,0 +1,259 @@
...
+        self.view1Decoder_ = _FullyConnectedNet(self.embedSize,
+                                                self.hiddenSize,
+                                                self.numHiddenLayers,
+                                                view1.shape[1]).to(device)
+        self.view2Decoder_ = _FullyConnectedNet(self.embedSize,
+                                                self.hiddenSize,
+                                                self.numHiddenLayers,
+                                                view2.shape[1]).to(device)
+
+        self.view1_encoder_ = self.view1Encoder_
```
bdpedigo: why?

tests/embed/test_splitae.py
cameronfr: We're not requiring pycodestyle for our tests. Imo enabling line-wrap in the editor instead of manually formatting is the way to go, hence the couple long lines here. 

I'm not sure if our line length requirements (80 our so characters) in e.g. splitae.py really made it easier to read. It definitely made it harder to edit, with all the new line breaks and the lining up of stuff horizontally.

tests/embed/test_splitae.py
cameronfr: Maybe we should save the data if the test fails because a failure on random data might reveal an edge case where the algorithm performs especially poorly. Or maybe that might just reveal data which is inherently impossible to cluster? Not sure.

cameronfr:
I don't think it's worth focusing so much attention on my lack of snake_case. The readability of the code does not change at all. Plus it seems much easier to type variable names without underscores especially when they're more descriptive (which I think is important).

PR #42, opened 2019-12-05
Multi-view Spectral Clustering

achang63:
An implementation of Multi-view Spectral Clustering based on the Co-training framework.


multiview/cluster/mv_spectral.py
```diff
@@ -0,0 +1,251 @@
...
+                self._info_view = info_view
+            else:
+                msg = 'info_view must be an integer between 0 and n_clusters-1'
+                raise ValueError(msg)
+
+        if not (isinstance(n_iter, int) and (n_iter > 0)):
+            msg = 'n_iter must be a positive integer'
+            raise ValueError(msg)
+
+        self._n_clusters = n_clusters
```
rflperry: underscores should be trailing

achang63: How closely are we following the Scikit-learn API? I just double-checked their docs and it seems that by their conventions, class fields that are set by parameters generally do not have a trailing underscore, while those set by the fit method do have trailing underscores.

rflperry: I guess if that's what they say, up to you. Just leading underscore variables are "hidden" so if someone may want to access them they should probably not have a leading underscore. Up to you

bdpedigo:
MultiviewSpectralClustering_Tutorial.ipynb
- Remove call to help
- need way more explanation of what is happening for a tutorial. Explain what problem the algorithm solves at a high level, at the very least. 
- I would consider using a simulated dataset also, and showing what happens with MV spectral vs regular spectral on concatenated features of the two views

MultiviewSpectralClustering_Validation.ipynb
- The point of this notebook is to call the code that you have merged into multiview, and run that code, not to reimplement the code in the notebook. Otherwise we don't know whether the code that is actually integrated really works
- I don't understand what is going on in the plots of the data. I would do the following in one single plot with subplots:
   - rows: view 1 - view 3
   - columns: col one is true data colored by true cluster label, col 2 is their implementation's results, col 3 is your results
   - in the above, remove all ticks and ticklabels, they dont add anything
- I'd like to see more validation in terms of simulations, I think it would be easy to run your algorithm on the same simulated data examples that @gmischl1 made. He has some really nice negative controls that I think would work well for you too.

multiview/cluster/mv_spectral.py
```diff
@@ -0,0 +1,232 @@
...
+            msg = 'max_iter must be a positive integer'
+            raise ValueError(msg)
+
+        self.n_clusters = n_clusters
+        self.n_views = n_views
+        self.random_state = random_state
+        self.info_view = info_view
+        self.max_iter = n_iter
+
+    def _gaussian_sim(self, X):
```
bdpedigo: just a note - i realize this is what was done in the paper - but gaussian similarity kernel is fairly restrictive, there are many other kernels one can use. Also, this is already implemented in sklearn, so I don't see why you need to implement it here (more to test, higher chance of bugs). 

I won't ask for it for the final for this semester, but I think addressing both of these comments would be worthwhile in the future. Basically the goal would be make yours have more of the options that sklearn's spectral clustering has

bdpedigo: also should be able to pick the bandwidth here

multiview/cluster/mv_spectral.py
```diff
@@ -0,0 +1,232 @@
...
+        '''
+
+        distances = cdist(X, X)
+        sq_dists = np.square(distances)
+        norm_dists = sq_dists / (2 * np.median(sq_dists))
+        sims = np.exp(-norm_dists)
+
+        return sims
+
+    def _compute_eigs(self, X):
```
bdpedigo: 1. there must be something in sklearn that already does this entire thing, like i mentioned early in the semester
2. does mv not have a consistent way of doing an embedding? Since you are doing eigendecomp of symmetric PSD matrix, can just use svd, as do many functions in multiview, so I see this as another place where you should either be able to recycle other code from the package or sklearn

multiview/cluster/mv_spectral.py
```diff
@@ -0,0 +1,232 @@
...
+                          for sim in new_sims]
+
+        # Row normalize
+        for view in range(self.n_views):
+            U_norm = np.linalg.norm(U_mats[view], axis=1).reshape((-1, 1))
+            U_norm[U_norm == 0] = 1
+            U_mats[view] /= U_norm
+
+        # Performing kmeans clustering
+        kmeans = KMeans(n_clusters=self.n_clusters,
```
bdpedigo: this is another thing that doesn't necessarily need to be `KMeans` in the future. 

also there are more params one could pass down to `KMeans` - for example, I almost always increase `n_init` from the default and find that it really helps. Consider allowing parameters to be passed down to `KMeans` or even pass down a different clustering algorithm

gmischl1:
As Pedigo said, feel free to take some of the simulated tests I made in the ctclassifier validation notebook. https://github.com/NeuroDataDesign/multiview/blob/CTClassifier2/docs/tutorials/cotraining/cotraining_classification_simulatedperformance.ipynb

multiview/cluster/mv_spectral.py
achang63: I replaced the portion involving calculating eigenvectors, but I still need most of that function to compute the normalized graph laplacian.

Issue #37, opened 2019-11-23
Implement CoTraining Classification


gmischl1:
Cotraining for classification tasks is important because it can be used in a wide variety of semi-supervised tasks, and it can be used in multiview settings or single view settings. Based on the paper by Blum and Mitchell, it is a very valuable algorithm which should be included in the multiview package.

To be most applicable, it should work with a variety of classifier types so that the user can specify many parameters about the classification process on their own.

References
Blum, Avrim, and Tom Mitchell. "Combining labeled and unlabeled data with co-training." Proceedings of the eleventh annual conference on Computational learning theory. ACM, 1998.

PR #44, opened 2019-12-12
CoTraining Classifier v2.0

gmischl1:
This PR updates the CTClassifier per Pedigo's comments. The main change to CTClassifier is to make it agree with the sklearn API, where all parameters are passed to __init__() rather than fit(). This PR also adds a full simulated testing notebook, as well as new unit tests for the classifier. It is meant to solve https://github.com/NeuroDataDesign/multiview/issues/37.


multiview/cotraining/base.py
```diff
@@ -60,15 +64,17 @@ def __init__(self,
     def fit(self, Xs, y):
         """
         A method to co-trained estimators to multiview data.
+
         Parameters
         ----------
-        Xs: list of array-likes
+        Xs: numpy.ndarray, list of array-likes
```
rflperry: For clarity I would have an "or", not a comma

multiview/cotraining/base.py
```diff
@@ -60,15 +64,17 @@ def __init__(self,
...
             - Xs shape: (n_views,)
```
rflperry: I would say Xs length: n_views

multiview/cotraining/ctclassifier.py
```diff
@@ -108,32 +143,62 @@ def __init__(
...
-            num_iter=50
+            y
             ):
         """
         Fit the classifier object to the data in Xs, y.
 
         Parameters
         ----------
-        Xs : list of array-likes
+        Xs : numpy.ndarray, list of array-likes
```
rflperry: same as above

multiview/cotraining/base.py
```diff
@@ -60,15 +64,17 @@ def __init__(self,
...
             - Xs[i] shape: (n_samples, n_features_i)
         y : array, shape (n_samples,)
+
         Returns
         -------
-        y_pred : array-like (n_samples,)
+        self : object
```
rflperry: sklearn says "returns an instance of self." instead of "object". not necessarily necessary but just saying

multiview/cotraining/ctclassifier.py
```diff
@@ -288,16 +317,19 @@ def fit(
...
+        return self
+
     def predict(self, Xs):
         """
         Predict the classes of the examples in the two input views.
 
         Parameters
         ----------
-        Xs : list of array-likes
+        Xs : numpy.ndarray, list of array-likes
```
rflperry: same as above

multiview/cotraining/ctclassifier.py
```diff
@@ -344,10 +376,11 @@ def predict_proba(self, Xs):
...
```
rflperry: same as above

multiview/cotraining/base.py
gmischl1: Ok. I changed this in ctclassifier and base.py to match your updated docstring format

PR #47, opened 2019-12-13
Update base class

rflperry:


mvlearn/embed/gcca.py
```diff
@@ -270,23 +272,3 @@ def transform(self, Xs, view_idx=None):
                     for x, proj in zip(Xs, self.projection_mats_)
                 ]
             )
-
-    def fit_transform(self, Xs):
-        """
```
gmischl1: Why are you removing the fit_transform functionality?

rflperry: Because it is implemented in the updated base class and GCCA doesn't change any aspect of it.

PR #49, opened 2019-12-18
Multiview KMeans Clustering

achang63:
This pull request includes the multi-view clustering algorithm, which is an extension of the kmeans clustering algorithm to multiple views using the co-training framework. The PR also includes unit tests, validation notebooks, and tutorial notebooks for the algorithm.


multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Xs = check_Xs(Xs, enforce_views=2)
+
+        # Run multi-view kmeans for n_init different centroid initializations
+        min_inertia = np.inf
+        best_centroids = None
+
+        for _ in range(self.n_init):
+
+            # Random initialization of centroids
```
bdpedigo: just so you know, this is a huge disadvantage for your alg vs. initializing with k-means++ and in my experience matters a lot. not going to make you look into it now, but I'd make an issue at least

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if not (isinstance(n_init, int) and (n_init > 0)):
+            msg = 'n_init must be a nonnegative integer'
+            raise ValueError(msg)
+
+        self.n_clusters = n_clusters
+        self.random_state = random_state
+        self.patience = patience
+        self.n_init = n_init
+        self.centroids_ = None
```
bdpedigo: don't need this I don't think

multiview/cluster/mv_k_means.py
bdpedigo: also, consider in the future adding option for custom initialization

bdpedigo:
- I'd prefer to see a plot or two of the simulation settings that you are describing in the validation notebook.
- is there anything in the paper to benchmark against?

multiview/cluster/mv_k_means.py
```diff
@@ -0,0 +1,330 @@
...
+            raise AttributeError(msg)
+
+        dist1 = cdist(Xs[0], self.centroids_[0])
+        dist2 = cdist(Xs[1], self.centroids_[1])
+        dist_metric = dist1 + dist2
+        predictions = np.argmin(dist_metric, axis=1).flatten()
+
+        return predictions
+
+    def fit_predict(self, Xs):
```
bdpedigo: this should be in base I think

multiview/cluster/mv_k_means.py
achang63: Will do

multiview/cluster/mv_k_means.py
achang63: Do you mean the entire implementation for the fit_predict method should be in base class?

multiview/cluster/mv_k_means.py
achang63: Do you mean the centroids attribute? I think the centroids need to be saved as an attribute during training in order to use them during predictions.

PR #62, opened 2020-01-21
Removed tqdm from requirements

rflperry:
Used to be a requirement but was removed as far as I can tell.


rguo123:
I think @cameronfr is using tqdm somewhere in his split ae code. Make sure to remove the dependency from those files so tests and docs will pass.

PR #98, opened 2020-03-04
DCCA

### LINKED ISSUE #80: Add Deep CCA
Deep CCA is a useful technique for enhancing the results of CCA with deep networks in a 2 view situation. It is analagous to training a special data-driven kernel for a modified kernel CCA. For the first sprint this semester, my ultimate goal will be to implement DCCA for this package based on the original paper:

http://proceedings.mlr.press/v28/andrew13.pdf

To do this, I will benchmark existing implementations, such as the following, in order to compare these implementations to the results of the paper, and ultimately develop code either by drawing from these implementations or starting from scratch. I will compare benchmark these implementations against Table 1 of the paper, which provides correlation results from a 2-view modified MNIST dataset. Then, if they seem to function well, I will port them into mvlearn. If they don't, I will work on my own implementation and benchmark it against these and the paper.

Existing DCCA implementations:

https://github.com/Michaelvll/DeepCCA
https://github.com/VahidooX/DeepCCA

gmischl1:
#### Reference Issues/PRs
Fixes #80 

This PR adds DCCA functionality for 2 view data. DCCA can be thought of as similar to Kernel CCA, where the kernel functions for each view are learned by deep networks before the linear CCA transformation such that the correlation between the output dimensions is maximized.


gmischl1:
> missing "import torch" I believe?
> 
> image

How did you get this error? I can't reproduce it. If I just do 

>>> from mvlearn.embed.dcca import DCCA

in a jupyter notebook and then try to instantiate a DCCA object, I get no errors

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+        self.epoch_num = epoch_num
+        self.batch_size = batch_size
+        self.learning_rate = learning_rate
+        self.reg_par = reg_par
+        self.print_train_log_info = print_train_log_info
+        self.threshold = threshold
+
+        self.deep_model = DeepCCA(layer_sizes1, layer_sizes2, input_size1,
+                                  input_size2, outdim_size,
+                                  use_all_singular_values, device=device)
```
rflperry: how is it calling itself? I'm getting an error trying to run it "NameError: name 'DeepCCA' is not defined"

gmischl1: Maybe not the best naming, but the class is DCCA. It should import DeepCCA (a different class) from utils in the line 
from ..utils.utils import *

gmischl1: Do you have the updated utils.py file?

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+
+from .base import BaseEmbed
+import torch
+from ..utils.utils import *
+
+
+class DCCA(BaseEmbed):
+    r"""
+    An implementation of Deep Canonical Correlation Analysis [#1DCCA]_ with
+    PyTorch. It computes projections into a common subspace in order to
```
rguo123: Maybe include something here that differentiates it from CCA description.

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+    maximize the correlation between pairwise projections into the subspace
+    from two views of data.
+
+    Parameters
+    ----------
+    input_size1 : int (positive)
```
rguo123: Nitpick: Not sure what styling is used in rest of repo but i think input_size_1 is better styling. Whichever one you use, just keep consistent with rest of package. Same for rest of params

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+    print_train_log_info : boolean, default=False
+        Whether or not to print the logging info (training loss at each epoch)
+        when calling DCCA.fit().
+    threshold : float, default=1e-2
+        Threshold difference between successive iteration losses to define
+        convergence and stop training.
+
+    Attributes
+    ----------
+    input_size1 : int (positive)
```
rguo123: in sklearn and the previous semester, I thought we were adding an underscore either before or after attribute name?

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+    maximize the correlation between pairwise projections into the subspace
+    from two views of data.
+
+    Parameters
+    ----------
+    input_size1 : int (positive)
+        The dimensionality of the input vectors in view 1.
+    input_size2 : int (positive)
+        The dimensionality of the input vectors in view 2.
+    outdim_size : int (positive), default=2
```
rguo123: is this n_components in gcca? I would go over gcca code and try to match up the variable names as much as possible since these are related algorithms.

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+        Threshold difference between successive iteration losses to define
+        convergence and stop training.
+
+    Attributes
+    ----------
+    input_size1 : int (positive)
+        The dimensionality of the input vectors in view 1.
+    input_size2 : int (positive)
+        The dimensionality of the input vectors in view 2.
+    outdim_size : int (positive), default=2
```
rguo123: see above comment about n_components

mvlearn/embed/dcca.py
rguo123: Imo linear_cca, Deep_cca, MLPNet, cca loss should not be put into utils since they seem specific only to dcca. If these are functions/classes that will be used in other methods as well then it should be fine to keep in utils. Otherwise, I think it'd be better to either have a specific dcca utils file or just include it with dcca.py.

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+        """
+        Xs = check_Xs(Xs, multiview=True)  # ensure valid input
+        x1 = torch.DoubleTensor(Xs[0])
+        x2 = torch.DoubleTensor(Xs[1])
+        x1.to(self.device)
+        x2.to(self.device)
+
+        data_size = x1.size(0)
+
+        checkpoint = 'checkpoint.model'
```
rguo123: is this what pytorch is saving? maybe set it as an attribute name (doesn't have to be a parameter). That way user knows where/what the checkpoints are being named in case fit() crashes.

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+
+        train_losses = []
+        epoch = 0
+        current_loss = np.inf
+        train_loss = 1
+        while (current_loss - train_loss > self.threshold)\
+                and epoch < self.epoch_num:
+            self.model.train()
+            batch_idxs = list(BatchSampler(RandomSampler(range(data_size)),
```
rguo123: Is this correct practice to create a data sampler at every iteration? Don't know any pytorch specifics so please lmk. Is there a good way of setting the seed for RandomSampler? Might be useful for replicating results (Obviously, be careful that you aren't setting the seed to the same value at each training epoch).

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+        self.threshold = threshold
+
+        self.deep_model = DeepCCA(layer_sizes1, layer_sizes2, input_size1,
+                                  input_size2, outdim_size,
+                                  use_all_singular_values, device=device)
+        self.linear_cca = linear_cca()
+
+        self.model = nn.DataParallel(self.deep_model)
+        self.model.to(device)
+        self.loss = self.deep_model.loss
```
rguo123: you should probably include the loss function as part of the class unless it's needed in other methods. That is, don't include it in utils but include it in either specific dcca_utils or dcca class itself.

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+        x2 = torch.DoubleTensor(Xs[1])
+
+        with torch.no_grad():
+            losses, outputs = self._get_outputs(x1, x2)
+            outputs = self.linear_cca.test(outputs[0], outputs[1])
+            if return_loss:
+                return outputs, np.mean(losses)
+            return outputs
+
+    def _train_linear_cca(self, x1, x2):
```
rguo123: Is this function necessary if it's only one line? You also need to document x1 and x2

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,349 @@
...
+        """
+        Private function to fit the linear CCA model for use after the
+        deep layers.
+        """
+        self.linear_cca.fit(x1, x2, self.outdim_size)
+
+    def _get_outputs(self, x1, x2):
+        """
```
rguo123: Missing param docs for x1 and x2. Also missing return docs

mvlearn/utils/utils.py
```diff
@@ -76,7 +93,7 @@ def check_Xs_y(Xs, y, multiview=False, enforce_views=None):
 
     Parameters
     ----------
-    Xs : nd-array, list
+    Xs : np-array, list
```
rguo123: everywhere else you are still using nd array? Is there a difference?

mvlearn/utils/utils.py
```diff
@@ -155,3 +172,345 @@ def check_Xs_y_nan_allowed(
              found {}".format(num_classes, n_classes))
 
     return Xs_converted, y_converted
+
+
+class linear_cca():
```
rguo123: How is this different from kcca with a linear kernel or gcca with 2 views? I would avoid replicating existing code (even tho at this point it'll take longer for you to change your code to use an existing implementation than keep as is).

rguo123: fyi going to skip code review for this class until this comment is resolved.

mvlearn/utils/utils.py
```diff
@@ -155,3 +172,345 @@ def check_Xs_y_nan_allowed(
...
+        result : nd-array
+            Result of linear transformation on input data.
+        """
+        result = x - self.m[idx].reshape([1, -1]).repeat(len(x), axis=0)
+        result = np.dot(result, self.w[idx])
+        return result
+
+    def test(self, H1, H2):
+        """
+        Transform inputs based on already fit matrices.
```
rguo123: why is this method called test instead of transform?

mvlearn/utils/utils.py
```diff
@@ -155,3 +172,345 @@ def check_Xs_y_nan_allowed(
...
+        H1, H2 = H1.t(), H2.t()
+
+        o1 = o2 = H1.size(0)
+
+        m = H1.size(1)
+
+        H1bar = H1 - H1.mean(dim=1).unsqueeze(dim=1)
+        H2bar = H2 - H2.mean(dim=1).unsqueeze(dim=1)
+
+        SigmaHat12 = (1.0 / (m - 1)) * torch.matmul(H1bar, H2bar.t())
```
rguo123: A lot of math going on here. Can you provide more comments on what the loss function is either here or in the class documentation?

mvlearn/embed/dcca.py
rflperry: I think it should be fine to include it in the dcca.py file

mvlearn/utils/utils.py
gmischl1: The new kcca hasn't been finished yet, so I don't have the code for that. This is the implementation of CCA that I got from the github that this DCCA is based off of. Once we have the updated KCCA pushed in, then I can switch to that, but until then I thought we could just use this basic CCA with no kernel.

mvlearn/utils/utils.py
gmischl1: Added extensive documentation in the Notes section of DCCA which will render online, and added a few line-by-line comments briefly explaining code.

mvlearn/embed/dcca.py
gmischl1: I switched it to be in dcca.py now

mvlearn/utils/utils.py
gmischl1: Documentation of DCCA includes extensive explanation, and small comments in line

rguo123:
Add an issue for removing linear_cca code in the future and adapting it with kcca if you haven't already?

rguo123:
@rflperry @gmischl1 
Github's not letting me respond to previous thread about utils.py. But I think it's better to create a dcca_utils.py since dcca.py is not close to a 1000 lines. At the minimum, make this an issue if you are close to deliverable deadline?

mvlearn/embed/dcca.py
rflperry: Don't think we have a standard but I asked in slack. Probably good to be consistent and use what most have been using thus far.

mvlearn/embed/dcca.py
```diff
@@ -0,0 +1,909 @@
...
+    loss_ : ``cca_loss`` object
+        Loss function for ``deep_model``. Defined as the negative correlation
+        between outputs of transformed views.
+    optimizer_ : torch.optim.RMSprop object
+        Optimizer used to train the networks.
+    tolerance_ : float, (positive)
+        Threshold difference between successive iteration losses to define
+        convergence and stop training.
+    is_fit_ : boolean
+        Whether or not .fit() has been called yet. Permits .transform() to
```
rflperry: being picky but wrap .fit() with `` ``

mvlearn/embed/dcca.py
rflperry: are checkpoints saved? might be useful. can just make an issue for it tho

rflperry:
Comments are in but seem nitpicky at this point. Squash and merge at your convenience. Any future potential edits/features make an issue for so they are noted.

gmischl1:
> Add an issue for removing linear_cca code in the future and adapting it with kcca if you haven't already?

Added an issue #101

mvlearn/embed/dcca.py
gmischl1: done

PR #106, opened 2020-03-25
Jive

akoul1:


mvlearn/jive/AJIVE.py
```diff
@@ -0,0 +1,776 @@
...
+from .random_direction import sample_randdir
+from .diagnostic_plot import plot_joint_diagnostic
+from .PCA import PCA
+
+
+class AJIVE(object):
+    """
+    An implementation of Angle-based Joint and Individual Variation Explained.
+    This algorithm 
+
```
gmischl1: Is there supposed to be more description here?

docs/tutorials/embed/mvmds_tutorial.ipynb
```diff
@@ -37,12 +37,24 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 8,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
rflperry, line 42: Seems to be an error in this notebook. Not sure you wanted to add this notebook in the PR?

mvlearn/__init__.py
```diff
@@ -4,3 +4,6 @@
 import mvlearn.cotraining
 import mvlearn.datasets
 import mvlearn.utils
+import mvlearn.jive
```
rflperry, line 7: It appears here and other files that the newline at the end of the file is being removed. This will result in a style check fail as far as I know. Also, in master the version is 0.0.2

tests/JIVE/test_AJIVE_real.py
```diff
@@ -0,0 +1,184 @@
+# -*- coding: utf-8 -*-
```
rflperry, line 1: I see a test_AJIVE and test_AJIVE_real. only want one id assume

rflperry:
Not sure the PR status but there are some branch conflicts and also I noticed folders named "JIVE". everything should be AJIVE or ajive for correctness.

mvlearn/__init__.py
akoul1: So should I take out this line?

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+        Number of random direction samples to draw.
+
+    precomp_randdir_samples {None,  array-like}
+        Precomputed random direction samples.
+
+    n_jobs: int, None
+        Number of jobs for parallel processing wedin samples and random
+        direction samples using sklearn.externals.joblib.Parallel.
+        If None, will not use parallel processing.
+
```
gmischl1, line 84: I'm confused by the 'type' descriptions of each parameter. Can you put them all in the format like
int, default=100, rather than {bool, None}. Maybe I'm just confused

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+        Parameters
+        ----------
+        blocks: {list of array-likes or numpy.ndarray, dict}
+            - blocks length: n_views
+            - blocks[i] shape: (n_samples, n_features_i) 
+            The different views that are input. Input as data matrices. 
+            If dict, will name blocks by keys, otherwise blocks are named by 
+            0, 1, ...K. 
+
```
gmischl1, line 189: Is there a specific reason this is allowed to be a dict, not just a list of array-likes? Does it have something to do with the way ajive is typically used in practice that the input would like be formatted as a dict already?

akoul1: Then return is as a dictionary as there are 3 different matrices returned. Also usually inputs are views with specific names. The keys make it easy to call those views and specify which of the 3 matrices you want

gmischl1:
Maybe it just wasn't rendering right when I looked, but it seemed like there were some issues in the AJIVE Tutorial.ipynb

In one of the cells at the top, it looked like a bunch of import statements were being formatted as text
"
import numpy as np import pandas as pd import seaborn as sns import matplotlib.pyplot as plt from mvlearn.jive import AJIVE from mvlearn.jive import PCA from IPython.display import Image"

mvlearn/ajive/ajive.py
gmischl1: Ok that's fine then, is there another way you can write the "type" of each input in the Parameters section though? I was very confused looking at {list, dict} the first time

mvlearn/ajive/ajive.py
gmischl1: I think something like list or dict, default=<DEFAULT> is more clear, but maybe that's just me

mvlearn/__init__.py
```diff
@@ -4,3 +4,6 @@
 import mvlearn.cotraining
 import mvlearn.datasets
 import mvlearn.utils
+import mvlearn.ajive
+
+__version__ = "0.0.1"
```
gmischl1, line 9: can you change the version to 0.0.2 so it's not conflicting with master?

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+    indiv_ranks: {list, dict, None}
+        Ranks of individual variation matrices. If None, will estimate the 
+        individual ranks. Otherwise, will use provided individual ranks.
+
+    center: {bool, None}
+        Centers matrix. If None, will not center.
+
+    reconsider_joint_components: bool
+        Triggers _reconsider_joint_components function
```
gmischl1, line 58: can you give a little more description of what this will do?

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+
+    joint_rank: {None, int}
+        Rank of the joint variation matrix. If None, will estimate the 
+        joint rank. Otherwise, will use provided joint rank.
+
+    indiv_ranks: {list, dict, None}
+        Ranks of individual variation matrices. If None, will estimate the 
+        individual ranks. Otherwise, will use provided individual ranks.
+
+    center: {bool, None}
```
gmischl1, line 54: change things like this to "bool, default=None" but also why is the default not just False? I think that would be less confusing.

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+        return results
+
+    def get_ranks(self):
+        """
+        Output
+        ------
+        joint_rank (int): the joint rank
+
+        indiv_ranks (dict): the individual ranks.
+        """
```
gmischl1, line 501: These are not formatted correctly. Should be

joint_rank : int
    The joint rank.
indiv_ranks : dict
    The individual ranks.

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+        Observation names.
+
+    var_names: None, array-like
+        Variable names for this view.
+
+    block_name: None, int, str
+        Name of this view.
+
+    m: None, array-like
+        The vector used to column mean center this view.
```
gmischl1, line 664: if the default of m is None, then this should say 
m : array-like, default=None

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+        The noise matrix estimate.
+
+    obs_names: None, array-like
+        Observation names.
+
+    var_names: None, array-like
+        Variable names for this view.
+
+    block_name: None, int, str
+        Name of this view.
```
gmischl1, line 661: See comment below about default being None, and likewise for above parameters.

docs/tutorials/embed/mvmds_proof.ipynb
```diff
@@ -22,8 +22,8 @@
    "metadata": {},
```
gmischl1, line 22: why is mvmds_proof being changed in this PR?

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import warnings
+from scipy.sparse import issparse
+from copy import deepcopy
+from sklearn.externals.joblib import load, dump
+import pandas as pd
+from mvlearn.embed.base import BaseEmbed
```
gmischl1, line 21: if ajive doesn't inherit from BaseEmbed, then no reason to import it

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+                _arg_checker(blocks,
+                            self.init_signal_ranks,
+                            self.joint_rank,
+                            self.indiv_ranks,
+                            precomp_init_svd,
+                            self.center)
+
+        block_names = list(blocks.keys())
+        num_obs = list(blocks.values())[0].shape[0] #number of views
+
```
gmischl1, line 207: you import check_Xs, but you don't use it. Is there any way to validate the input you are getting? For example, if you need there to be exactly 2 views, check_Xs can be used to do that if you first convert the blocks to a temporary list of matrices

docs/tutorials/embed/mvmds_proof.ipynb
akoul1: Accidental haha. Going to change back

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+                                                   individual=block_specific\
+                                                   [bn]['individual'],
+                                                   noise=block_specific[bn]\
+                                                   ['noise'],
+                                                   block_name=bn,
+                                                   obs_names=obs_names,
+                                                   var_names=var_names[bn],
+                                                   m=self.centers_[bn],
+                                                   shape=blocks[bn].shape)
+
```
gmischl1, line 422: can you add an attribute called self.is_fit_ which is boolean and set to False in the __init__() function, but set to True when fit() is done, and then this attribute is checked in other methods that return things only if fit() has been called already? If it hasn't return some error message like "Must call fit() first."

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+    individual: mvlearn.ajive.pca.pca
+        view specific individual PCA.
+        Has an extra attribute individual.full_ which contains the full view
+        joint estimate.
+
+    noise: array-like
+        The full noise view estimate.
+
+    block_name:
+        Name of this view.
```
gmischl1, line 683: is this a string? need to list its type

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,760 @@
...
+    def __init__(
+        self,
+        joint,
+        individual,
+        noise,
+        obs_names=None,
+        var_names=None,
+        block_name=None,
+        m=None,
+        shape=None,
```
gmischl1, line 696: shape can be passed in, but it doesn't appear to be documented in the Parameters section above

gmischl1:
In the tutorial notebook, you use a heatmap visualization that you define within the notebook. Is this instead of using the plotting functions defined in the files like block_visualization? Can you use some of the visualization tools that are in the package to plot stuff instead of manually defining a heatmap function?

gmischl1:
Overall, my main comment is to clean up the documentation. A lot of the Parameters or Outputs sections were missing things or weren't formatting uniformly. Also, fix a couple things in the tutorial notebook

mvlearn/__init__.py
rflperry: You want a newline at the end of every file (I believe) else style check will fail.

gmischl1:
You'll need to add a docs/references/ajive.rst file (following the examples in that folder, like cluster.rst) so it shows up in the documentation, as well as add the path to the tutorial notebook to the docs/tutorials.rst file.

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,776 @@
...
+        Percentile for random direction (lower) bound cutoff for squared
+        singular values used to estimate joint rank.
+
+    n_randdir_samples: int = 1000
+        Number of random direction samples to draw.
+
+    precomp_randdir_samples: array-like = None
+        Precomputed random direction samples.
+
+    n_jobs: int = None
```
gmischl1, line 79: To give a default, these should all say "int, default=None", not "int = None". Sorry, but you should change this for all your methods if you did it this way

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,776 @@
...
+
+    @classmethod
+    def load(cls, fpath):
+        return load(fpath)
+
+    def get_full_block_estimates(self):
+        """
+
+        Output
+        ------
```
gmischl1, line 449: Returns, not Output

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,776 @@
...
+            full[bn] = {'joint': self.blocks[bn].joint.full_,
+                        'individual': self.blocks[bn].individual.full_,
+                        'noise': self.blocks[bn].noise_}
+
+        return full
+
+    def results_dict(self):
+        """
+        Returns all estimates as a dicts.
+
```
gmischl1, line 465: If this is a method that users are allowed to call (not just a private method for within the class), then it needs complete documentation. Same goes for all methods that aren't private

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,776 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import warnings
+from scipy.sparse import issparse
+from copy import deepcopy
+from sklearn.externals.joblib import load, dump
+import pandas as pd
+from mvlearn.utils.utils import check_Xs
```
gmischl1, line 21: don't import from mvlearn, do relative imports, like "from ..utils.utils import check_Xs"

mvlearn/ajive/ajive.py
```diff
@@ -0,0 +1,776 @@
...
+    joint_rank_wedin_est_: int
+        The joint rank estimated using the wedin/random direction bound.
+
+    joint_rank: int
+        The rank of the joint matrix
+
+    indiv_ranks: dict of ints
+        Ranks of the individual matrices for each view.
+
+    """
```
gmischl1, line 127: Definitely should have a References section

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,937 @@
...
+from .ajive_utils.wedin_bound import get_wedin_samples
+from .ajive_utils.random_direction import sample_randdir
+from .ajive_utils.pca import pca
+
+
+class ajive(object):
+    r"""
+    An implementation of Angle-based Joint and Individual Variation Explained.
+    This algorithm takes multiple input views with the same number of samples
+    and decomposes them into 3 distinct matrices representing [#1ajive]_:
```
rflperry, line 33: Rephrased. All multiview algorithms require same number of samples. Pulled some details from the JIVE paper abstract, added a note that I just added for GCCA too.

An implementation of Angle-based Joint and Individual Variation Explained  [#1ajive]_.
This algorithm takes multiple views and decomposes them into 3 distinct matrices representing:
    - Low rank approximation of individual variation within each view
    - Low rank approximation of joint variation between views
    - Residual noise
An important note, AJIVE can handle any number of views, not just two.

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,937 @@
...
+        - Individual variation of each particular view
+        - Joint variation shared by all views
+        - Noise
+
+    Parameters
+    ----------
+    init_signal_ranks: list or dict
+        The initial signal ranks.
```
rflperry, line 41: Bit confused to what this is, what are the signal ranks?

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+    n_randdir_samples: int, default = 1000
+        Number of random direction samples to draw.
+
+    precomp_randdir_samples: array-like, default = None
+        Precomputed random direction samples.
+
+    Attributes
+    ----------
+
+    common: mvlearn.factorization.pca.pca
```
rflperry, line 81: This is a class? not like a matrix? Also this and block should technically end with underscores.

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+    wedin_cutoff_: float
+        Singular value squared cutoff for the wedin bound.
+
+    svalsq_cutoff_: float
+        max(rand_cutoff_, wedin_cutoff_)
+
+    joint_rank_wedin_est_: int
+        The joint rank estimated using the wedin/random direction bound.
+
+    joint_rank: int
```
rflperry, line 115: Should end with underscores

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+    :math:`X^{i}` matrix onto the orthogonal complement of :math:`U_{joint}`:
+
+        .. math::
+            X^{(i), orthog} = (I - U_{joint}U_{joint}^T)X^{(i)}
+
+
+    :math:`I` in the above equation represents the identity matrix.
+    From here, the :math:`I^{(i)}` matrix (individual signal estimate) can be
+    found by performing the rank :math:`r_{individual}^{(i)}` singular value
+    decomposition of :math:`X^{(k), orthog}`. :math:`r_{individual}^{(i)}` can
```
rflperry, line 183: should be :math:`X^{(i), orthog}` right?

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+        else:
+            return "No data has been fitted yet"
+
+    def fit(self, blocks, precomp_init_svd=None):
+        r"""
+        Fits the AJIVE decomposition.
+
+        Parameters
+        ----------
+        blocks: dict or list of array-likes
```
rflperry, line 277: blocks needs to be renamed to Xs. Also, I don't see a reason why dictionaries should be allowed as input? Just list of array-likes. Find replace should do it

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+                indiv_rank = self.blocks[bn].individual.rank
+                r += ", block {} indiv rank: {}".format(bn, indiv_rank)
+            return r
+
+        else:
+            return "No data has been fitted yet"
+
+    def fit(self, blocks, precomp_init_svd=None):
+        r"""
+        Fits the AJIVE decomposition.
```
rflperry, line 273: A little more of a description. What parameters is it learning exactly?

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+        for bn in self.block_names:
+            full[bn] = {'joint': self.blocks[bn].joint.full_,
+                        'individual': self.blocks[bn].individual.full_,
+                        'noise': self.blocks[bn].noise_}
+
+        return full
+
+    def results_dict(self):
+        r"""
+
```
rflperry, line 559: Sentence description here

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+        Returns
+        -------
+
+        results: dict of dict of dict of np.arrays
```
rflperry, line 563: Don't really like the idea of returning a dictionary. Would prefer a list of the three but maybe something I'll look at in terms of a larger reworking of AJIVE

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+                                          'loadings': indiv.loadings_,
+                                          'rank': indiv.rank,
+                                          'full': indiv.full_},
+
+                           'noise': self.blocks[bn].noise_}
+
+        return results
+
+    def get_ranks(self):
+        r"""
```
rflperry, line 609: description here

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+        if not self.is_fit:
+            raise ValueError("Decomposition has not yet been computed")
+
+        joint_rank = self.common.rank
+        indiv_ranks = {bn: self.blocks[bn].individual.rank for bn in
+                       self.block_names}
+        return joint_rank, indiv_ranks
+
+    def data_block_heatmaps(blocks):
+        r"""
```
rflperry, line 627: Description

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+
+        Returns
+        -------
+        fig : figure object
+            Figure returned contains the heatmaps of all views
+        """
+        _data_block_heatmaps(blocks)
+
+    def ajive_full_estimate_heatmaps(full_block_estimates, blocks):
+        r"""
```
rflperry, line 643: description

mvlearn/factorization/ajive.py
```diff
@@ -0,0 +1,921 @@
...
+
+    # remove columns of joint_scores that don't satisfy the constraint
+    joint_rank = len(to_keep)
+    joint_scores = joint_scores[:, list(to_keep)]
+    joint_loadings = joint_loadings[:, list(to_keep)]
+    joint_svals = joint_svals[list(to_keep)]
+    return joint_scores, joint_svals, joint_loadings, joint_rank
+
+
+class ViewSpecificResults(object):
```
rflperry, line 801: I think this makes more sense in the utils folder? It is never passed to the user, is just used for internal storage. It's going to appear in the webpage if here.

mvlearn/factorization/ajive.py
akoul1: Initially the ranks are essentially guessed. It's kind of weird. Basically you guess what the signal is apart from noise-like stuff (from the scree plot), then you find the joint basis, then you go and find the individual ranks

mvlearn/factorization/ajive.py
akoul1: I think I'm going to change .predict() to return a list of 3. I think the results_dict should stay bc there's a decent amount of other data that is returnable (scores, svals, etc.) and it's nicely formatted

mvlearn/factorization/ajive.py
akoul1: I didn't create a .rst module for it so when I built the HTML it didn't seem to be there. Do you think that it could still show up when merged? Either way, not a big deal to move to utils

mvlearn/factorization/ajive.py
akoul1: It's a pca object from the pca class

mvlearn/factorization/ajive.py
akoul1: Moved it to the pca file. Doesn't work in the utils file bc the pca file calls it (becomes circular)

mvlearn/factorization/ajive.py
akoul1: I changed it to a list but upon and tried it but when I ended up changing it, things got a bit confusing given you have a list of lists (if you want to find the individual matrix of the third view you would have to type full[2][1] which takes a decent amount of thought). I think a potential solution could be returning a pd.DataFrame. What do you think about that?

akoul1: that being said, it isn't hard to change it to return the list of lists for the .predict() function (or even have a boolean that returns one or the other)

akoul1:
> Quite the large files and PR but exciting stuff! Made a bunch of comments. I'm not well versed in the algorithm enough so it's hard to evaluate some things but I trust the code/tests/tutorial. General comments
> 
> 1. Documentation looks good in a lot of areas but there are some functions later on that need descriptions They're simple functions so a sentence will do but the functions will appear in the docstrings.
> 2. Saw the heatmap stuff is there. Great for the tutorial. I take it it is generalized enough for any data, not just the toy tutorial data?
> 3. Couple variable names to fix up, mainly underscores on some variables. Big one is it should be Xs, not blocks
> 4. Related to 3, lots of parameters take in/use dictionaries. That's not really part of our API so at least for the inputs, if the dicts can be removed that would be ideal. Just lists
> 5. The tutorial look fantastic overall! Not sure that the figure is needed, especially if it is in the paper but give me your thoughts. You did use the wrong "block" in the second set of plots, so there is still no individual variation. Easy fix. Also, did the paper say initial ranks set to 2? because the scree plots sort of indicate 3 to me (but it's all sort of a guessing game anyways I'd say).
> 
> 1. Related to 4, the use of dicts as a return not really part of our API and so I think should be reworked.
> 2. Some of the rank selection stuff could maybe be better integrated with some of the other rank selection tools
> 3. There's a lot of verbosity in the package. Functions and stuff that seem overcomplicated.
> 
> I've separated comments 3-6 because they are more stylistically oriented. but more complicated The first 3 are needed, much less complicated. Some of the stuff in the latter 3 (like 5) comes to mind because it mimics some of the GCCA work I've done. If you want to do the first 3, I'd look at the latter 3 and make some small progress with that (won't be able to get to it all but oh well). Notably, there's a rank selection method we use in GCCA that would be potentially useful here.

mvlearn/factorization/ajive.py
rflperry: So its sort of the sum of the ranks of the joint and individual embeddings? Maybe change to "Initial guess of the joint+individual rank for each view". I think I may set this up to use the rank selection method i told you about by default.

PR #109, opened 2020-03-30
Documentation Examples and import updates

### LINKED ISSUE #111: Add Examples Section to all Class Documentation
## Expected Behavior
To make documentation more helpful for people who want quick example usage without having to parse through long tutorial notebooks, have a quick example section in the docstring for each class.

## Actual Behavior
Currently, only a couple classes have this.

gmischl1:
#### Linked Issues
Fixes #111 

This PR adds an Examples section to the class documentation for every class except KCCA, since that is an active PR right now.  These render in sphinx and improve the documentation for users. 

Also, the import statements were changed in all the tutorial notebooks so that they follow this structure
```mvlearn
from mvlearn.module import ClassName
```
rather than
```mvlearn
from mvlearn.module.filename import ClassName
```

This PR also updates the MVMDS tutorial so that it includes a more direct comparison between it and PCA.


mvlearn/cotraining/ctclassifier.py
```diff
@@ -113,6 +119,40 @@ class CTClassifier(BaseCoTrainEstimator):
         The starting random seed for fit() and class operations, passed to
         numpy.random.seed().
 
+    Examples
```
rflperry, line 122: This seems long. Off of the top of my head I don't recall seeing any sklearn like documentation this long but could be wrong. Tutorial vs this tradeoff. Up to you.

rflperry: Also, I assume that you compiled the docs and all of these look fine. There is a way for tests to run and compile these, I am not sure if we have that implemented in our travis though.

PR #112, opened 2020-03-30
Added shuffle option to UCI dataloader

gmischl1:
This adds the shuffle parameter to the uci dataloader so that data is not returned in its unshuffled state. By default, the parameter shuffle=False, and the data will be returned in a deterministic order. However, the data will no longer be returned in its ordered form. If shuffle is set to True, then the user can either pass in a given random_state argument to shuffle deterministically, or they can allow the code to randomly shuffle in a non-deterministic way.


mvlearn/datasets/base.py
```diff
@@ -25,9 +26,17 @@ def load_UCImultifeature(select_labeled="all"):
         specified, all examples in the dataset are returned. Repeated labels
         are ignored.
 
+    shuffle : bool, default=False
+        If ``True``, returns each array with its rows and corresponding
```
rflperry, line 30: I think you just need single \` marks for \`True\` to be `True`

rflperry: Same for random_state

mvlearn/datasets/base.py
```diff
@@ -44,6 +53,11 @@ def load_UCImultifeature(select_labeled="all"):
     if select_labeled == "all":
         select_labeled = range(10)
 
+    if not shuffle:
+        random_state = 1
+    elif random_state is None:
```
rflperry, line 58: I think that random_state being None for np.random.seed will invoke a random seeding without the need to randomly assign random_state. Effectively, `np.random.seed(None)` will yield a random seeding.

mvlearn/datasets/base.py
gmischl1: Single didn't work for me when I tried building the docs for some reason

PR #113, opened 2020-04-01
Synth data

rflperry:
Implements a data generator. The same latent variables are generated from gaussian mixtures (supervised) and then one view is transformed according to a given transformation (built in defaults but accepts a custom function).


mvlearn/utils/plotting.py
```diff
@@ -0,0 +1,49 @@
+import matplotlib.pyplot as plt
+import seaborn as sns
+from .utils import check_Xs_y
+from .utils import check_Xs
+
+
+def plot_2views(
+    Xs,
```
gmischl1, line 8: What are your thoughts on renaming this to something like "plot_2views_grid" or something that describes that it will do this grid of feature-wise plots? I could see us adding a subplot_nviews or something like that which actually plots each view separately (assuming they've been transformed to 2d), just to differentiate this from anything else in the future

gmischl1: Or like plot_2views_cross

rflperry: Yes I'd prefer a new name. It reminds me of seaborn's pairplot but across views instead of within one. One of the following, plus/minus the word plot?

- crossviewplot? 
- twix_viewplot? 
- **crossplot**??? 
- pair2plot?
- pairedviews_plot?
- crossviews_plot?
- cross2views_plot?

I like crossplot personally but maybe because it's short and thus too vague.

gmischl1: I like pairedviews_plot the most, then cross2views_plot or crossviews_plot

theohj826: duoviews_plot? i dont really like the 2 in the function name. crossviews_plot sounds good too

mvlearn/utils/plotting.py
```diff
@@ -0,0 +1,49 @@
...
+    r"""
+
+    """
+    if y is None:
+        Xs = check_Xs(Xs)
+    else:
+        Xs,y = check_Xs_y(Xs, y)
+
+    n = Xs[0].shape[1]
+    fig, axes = plt.subplots(n, n, figsize=figsize, **fig_kwargs)
```
gmischl1, line 27: While I think the default should this, which is a grid of figures for every dimension in X, I think it would be great if there were an option for the user to input a list of indices that they want to plot (so like if they have high dimensional data but they know only that features 0, 1, and 5 are potentially related, they could input an optional list [0, 1, 5] and then this would only make a 3x3 grid plotting those features against each other. Cuz if people have high dimensional data this plotting function won't be as useful

rflperry: Added, good point

mvlearn/utils/plotting.py
rflperry: I settled on crossviews_plot

rflperry:
Plotting and simulator are up. Html documentation looks good and I made some minor updates to some HTML to make headers look better. (mvlearn.embed => Embed)

mvlearn/plotting/plot.py
```diff
@@ -0,0 +1,144 @@
...
+        - Xs length: n_views
+        - Xs[i] shape: (n_samples, n_features_i)
+        The two views to plot against one another. If one view has fewer
+        dimensions than the other, only that many will be plotted.
+    labels : boolean, default=None
+        Sets the labels of the samples.
+    dimensions : array-like of ints, default=None
+        The dimensions of the views to plot. If `None`, all dimensions up
+        to the minimum bettern the views will be plotted.
+    figsize : tuple, default=(10,10)
```
gmischl1, line 67: Typo "bettern" should be "between" I think

mvlearn/datasets/gaussian_mixture.py
```diff
@@ -0,0 +1,207 @@
...
+            The variance(s) of the multivariate Gaussian(s). If `class_probs`
+            is None (default), then is a 2D array-like of the multivariate
+            variance. Otherwise is a list of variances of the same length as
+            `class_probs`.
+        n : int
+            The number of samples to sample.
+        class_probs : array-like, default=None
+            A list correponding to the fraction of samples from each class and
+            whose entries sum to 1. If `None`, then data is sampled from a one
+            class.
```
gmischl1, line 98: change to "from one class" not "from a one class"

mvlearn/__init__.py
```diff
@@ -4,5 +4,6 @@
 import mvlearn.cotraining
 import mvlearn.datasets
 import mvlearn.utils
+import mvlearn.plotting
 
 __version__ = "0.0.1"
```
gmischl1, line 9: from what I remember, this version becomes the version number for the thing published to PyPi, should we change to 0.0.2 or something higher before updating the package? Your call

rflperry: yeah I guess this is all part of the next version, so should do it.

PR #121, opened 2020-04-07
Quick visualize plotting function

gmischl1:
This PR adds the quick_visualize plotting function, which performs MVMDS with 2 components on multi-view data and plots over 2 dimensions so that users can quickly get a sense of what their data looks like. 

#### Any other comments?
Also, 2 png files are added to the figures folder and are compiled in the docs for the plotting functions..


mvlearn/plotting/plot.py
```diff
@@ -68,6 +69,8 @@ def crossviews_plot(
         Sets the grid figure size.
     title : string, default=None
         Sets the title of the grid.
+    cmap : String, default=None
```
rflperry, line 72: Good catch

mvlearn/plotting/plot.py
```diff
@@ -135,10 +148,116 @@ def crossviews_plot(
 
     if title is not None:
         plt.tight_layout(rect=[0, 0.03, 1, 0.95])
-        plt.suptitle(title)
+        ax.set_title(title)
```
rflperry, line 151: Is this the same thing? before it was one title for all the plots and I recall it working all okay. Did something break?

mvlearn/plotting/plot.py
```diff
@@ -135,10 +148,116 @@ def crossviews_plot(
...
+    if ax_labels:
+        plt.xlabel("Component 1")
+        plt.ylabel("Component 2")
+    if not ax_ticks:
+        plt.xticks([], [])
+        plt.yticks([], [])
+
+    if title is not None:
+        plt.tight_layout(rect=[0, 0.03, 1, 0.95])
+        plt.title(title)
```
rflperry, line 257: plt vs ax? title vs. suptitle? See above comment

mvlearn/plotting/plot.py
gmischl1: Oops, I didn't mean to change it for crossviews_plot, I meant to do it for quick_visualize. I'll change that back

mvlearn/plotting/plot.py
gmischl1: If it's only a single figure, rather than a grid, suptitle results in an off-center title. So I used plt.title here. ax.set_title resulted in some weird issues where axis labels were different sizes depending on the title

gmischl1:
Changed according to comments from @rflperry and added a tutorial notebook

Issue #69, opened 2020-01-24
Torch is fairly large to install. Consider making an optional dependency?


rflperry:

bdpedigo:
+1

PR #126, opened 2020-04-14
Optional torch

rflperry:
#### Reference Issues/PRs
#69 

Modifies installation to make torch (and dependencies for torch related files) optional

#### Any other comments?
Documentation for installation updated as well. See below

image

image


docs/install.rst
```diff
@@ -1,55 +1,69 @@
...
+
+with optional dependencies
+
 - torch
 - torchvision
 - tqdm
+- pillow
 
 Hardware requirements
 ---------------------
```
gmischl1, line 69: Can you put the single quotes around the "mvlearn" under Testing?

docs/install.rst
```diff
@@ -1,55 +1,69 @@
...
-Install the current release of ``mvlearn`` with ``pip``::
+Install the current release of ``mvlearn`` with ``pip3``::
 
-    $ pip install mvlearn
+    $ pip3 install mvlearn
 
 To upgrade to a newer release use the ``--upgrade`` flag::
 
-    $ pip install --upgrade mvlearn
+    $ pip3 install --upgrade mvlearn
```
gmischl1, line 24: How do you upgrade if you want the torch dependencies? Does this single command do it, or do you need something like 

pip install --upgrade mvlearn[torch]

gmischl1:
Just so I know, what happens if:

- Someone already installed mvlearn 0.0.1 which has SplitAE, and now decides to do 
```mvlearn
pip install --upgrade mvlearn
```
does this mean they can still use splitAE from the old version, or does that get removed for them?

- Someone has mvlearn 0.0.2 without torch, and wants to upgrade to 0.0.3 at some point in the future with torch. What command do they run?

gmischl1:
Can you add a line to the Notes section of the docstrings for DCCA and SplitAE that says something like

"this function is only available if mvlearn was installed with the optional torch dependencies" and then links to the Install page?

docs/install.rst
rflperry: see below. Tell me if refined documentation is needed.

rflperry:
> Just so I know, what happens if:
> 
> * Someone already installed mvlearn 0.0.1 which has SplitAE, and now decides to do
> 
> ```
> pip install --upgrade mvlearn
> ```
> 
> does this mean they can still use splitAE from the old version, or does that get removed for them?
> 
> * Someone has mvlearn 0.0.2 without torch, and wants to upgrade to 0.0.3 at some point in the future with torch. What command do they run?
> * Someone has mvlearn 0.0.2 without torch, but wants torch. Can they just do
> 
> ```
> pip3 install mvlearn[torch]
> ```

rflperry:
Key points:
```mvlearn
pip3 install mvlearn
```
and
```mvlearn
pip3 install mvlearn[torch]
```
differ (irrespective of an option flags) in that the latter also checks extra dependencies. Without [torch], the command will ignore those extras.

## Case 1:
### Scenario
- Someone already installed mvlearn 0.0.1 with torch and then decides to do
```mvlearn
pip3 install --upgrade mvlearn
```
### Answer
torch dependencies will remain the same as before but everything else will be updated. So, in our current case, SplitAE will still function.

## Case 2:
### Scenario
- Someone has mvlearn 0.0.2 without torch, and wants to upgrade to 0.0.3 at some point in the future with torch. What command do they run?

### Answer
```mvlearn
pip3 install --upgrade mvlearn[torch]
```
## Case 3:
### Scenario
- Someone has mvlearn 0.0.2 without torch, but wants torch. Can they just do 
```mvlearn
pip3 install mvlearn[torch]
```
### Answer
Yes

PR #127, opened 2020-04-15
Update links in README

gmischl1:
This PR fixes the links in the README so they no longer reference NeuroDataDesign github, but rather neurodata. This also fixes the ``build`` badge which previously said "unknown" but now says "passing"


rflperry:
Good catch, looks great.

PR #130, opened 2020-04-15
Kcca website fixes

theohj826:
Fixes tutorial.rst to route to new tutorial notebook and examples section in kcca.py file because it wasnt rendering before and made it shorter. Also made some small changes to tutorial notebook--> changed the training/testing ratio for linear kcca.


gmischl1:
Just one thing: in the tutorial notebook, in another markdown cell under each major header (like under the "Polynomial Kernel Implementation" and same for Gaussian), can you write something like

"Here we show how KCCA with a polynomial kernel can uncover the highly correlated latent distribution of the 2 views which are related with a polynomial relationship, and then transform the data into that latent space."

theohj826:
Thanks gavin, just added that to the tutorial notebook^

PR #132, opened 2020-04-15
Gaussian shuffle and tutorials

gmischl1:
This PR adds shuffle and random_state setting ability to the GaussianMixtures class for synthetic data generation. It also updates the gaussian mixture tutorial, and uses the gaussian mixtures in a dcca tutorial notebook.

#### Any other comments?
It also fixes minor documentation errors like missing Attributes section in GaussianMixtures


mvlearn/datasets/gaussian_mixture.py
```diff
@@ -139,6 +181,12 @@ def __init__(self, mu, sigma, n, class_probs=None):
                 ]
             )
 
+        if self.shuffle:
+            np.random.seed(self.shuffle_random_state)
```
rflperry, line 185: Correct me but I believe this would shuffle y and latent independently if shuffle_random_state is None.

rflperry: Shuffle a list of indices and apply to both

rflperry:
There appears to be a conflict too, with docs/tutorials.rst

mvlearn/datasets/gaussian_mixture.py
gmischl1: I also set the random seed before I shuffle the labels below though. I plotted both and I believe they are shuffled correctly.

gmischl1:
Fixed issue with seeding for shuffle.

PR #133, opened 2020-04-19
Fix import

rflperry:
Importing without torch fails because it attempts to import files which import torch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: python
 branches:
   only:
   - master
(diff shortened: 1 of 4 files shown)


gmischl1:
Is there a way you can generate a warning if the user tries to import SplitAE or DCCA without having torch?

rflperry:
updated with warning. Tell me if if looks good and I'll PR it. I think the travis is just weird because of the torch imports.

PR #135, opened 2020-04-21
Mvmds Doc Updates

akoul1:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,0 @@
-# Auto detect text files and perform LF normalization
-* text=auto
--- a/mvlearn/embed/mvmds.py
+++ b/mvlearn/embed/mvmds.py
@@ -66,6 +66,42 @@ class MVMDS(BaseEmbed):
     >>> print(Xs_reduced.shape)
     (2000, 5)
 
+    Notes
+    -----
+    Classical Multiview Multidimensional Scaling can be broken down into two
+    steps. The first involves finding the common principal components of the
+    matrices. These can be thought of as multiview generalizations of the 
+    principal components found in principal component analysis (PCA) given
+    several covariance matrices. The central hypothesis of the common principal
+    component model states that given k normal populations (views), their
+    :math:`p` x :math:`p` covariance matrices :math:`\Sigma_{i}`, for
+    :math:`i = 1,2,...,k` are simultaneously diagonalizable as:
+    
+    ..math::
+        \Sigma_{i} = QD_i^2Q^T
+        
+    where Q is the common :math:`p` x :math:`p` orthogonal matrix and 
+    :math:`D_i^2` are positive :math:`p` x :math:`p` diagonal matrices. This
+    algorithm solves the maximum likelihood estimate of :math:`Q` 
+    Consider two views :math:`X_1` and :math:`X_2`. Multiview Multidimensional
+    Scaling seeks to find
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
```
gmischl1, line 2: what is the reason for adding this?

mvlearn/embed/mvmds.py
```diff
@@ -49,6 +49,45 @@ class MVMDS(BaseEmbed):
...
+    problem:
+
+    .. math::
+        \text{Minimize} \sum_{i=1}^{k}n_ilog(q_j^TS_iq_j)
+    .. math::
+        \text{Subject to } q_j^Tq_j = 1
+
+    where :math:`n_i` represent the degrees of freedom and :math:`S_i`
+    represent sample covariance matrices.
+
```
rflperry, line 90: Is there an actual paper to add a citation for? Say https://arxiv.org/abs/1605.00286

mvlearn/embed/mvmds.py
```diff
@@ -49,6 +49,45 @@ class MVMDS(BaseEmbed):
...
+    Classical Multiview Multidimensional Scaling can be broken down into two
+    steps. The first step involves calculating the Euclidean Distance matrices, 
+    :math:`X_i`, for each of the :math:`k` views and double-centering 
+    these matrices through the following calculations:
+
+    .. math::
+        \Sigma_{i}=-\frac{1}{2}J_iX_iJ_i
+
+    .. math::
+        \text{where }J_i=I_i-{\frac {1}{n}}11_i'
```
rflperry, line 62: I think `11_i`  is a typo? you want `\mathbb{1}\mathbb{1}^T` I think. Also the choice of X_i is a bit misleading since the ith view data matrix we usually represent as X_i

mvlearn/embed/mvmds.py
akoul1: That's for metric MDS. This is classical mds which is effectively scaled pca

.gitattributes
akoul1: hmm I didn't add this. It seems like it was automatically added when I create the branch

mvlearn/embed/mvmds.py
akoul1: This is the same documentation as the following: https://en.wikipedia.org/wiki/Multidimensional_scaling

akoul1: Will change the X_i

mvlearn/embed/mvmds.py
rflperry: Sorry, I meant for MVMDS in general.

rflperry:
Pycodestyle errors blocking
```
$ pycodestyle mvlearn/
mvlearn/embed/mvmds.py:52:1: W293 blank line contains whitespace
mvlearn/embed/mvmds.py:54:80: W291 trailing whitespace
mvlearn/embed/mvmds.py:55:70: W291 trailing whitespace
The command "pycodestyle mvlearn/" exited with 1.
```

.gitattributes
gmischl1: Looking at the master branch, it seems like we actually don't have a .gitattributes file right now, so this would be adding a new file. So, if you can just delete this file from the PR alltogether, that should be fine

gmischl1:
Since you have a PR for MVMDS docs up anyway, can you also make this quick change?

In the Examples section, replace the lines
>>> print(Xs[0].shape[0])  # number of samples in each view
   2000

with these lines
>>> print(len(Xs))  # number of samples in each view
   6
>>> print(Xs[0].shape)  # number of samples in each view
   (2000, 76)

PR #142, opened 2020-04-26
Cluster fig tutorial

gmischl1:
This PR adds the full tutorial used to produce the clustering figures from the paper.


rflperry:
I like adding this. I think conciseness is something to seek in our tutorials. Is it possible to plot the views side by side? It currently takes up a lot of vertical space as is.

Is it possible to plot some of the others side by side as well? I think it would make comparisons easier and the notebook shorter. I ended up not using the plotting functionality in my cca comparison notebook so I could plot things together.

rflperry:
Also I would make clear headers, and the first header can be one '#' sign with other ones incrementing as need be

rflperry:
Seems to be a conflict with the updated tutorials.

PR #145, opened 2020-04-26
Tutorial fix

rflperry:
At least one broken tutorial and others with too lengthy titles. Also added some of the verification ones to be shown again.


gmischl1:
In the MVMDS vs PCA tutorial, it seems like there's something weird going on with the ari scores. For both the noiseless and noisy case, the score are exactly the same. Also, for the noisy case, the printout shows MVMDS having a better score, but it says pca performs better.

rflperry:
> In the MVMDS vs PCA tutorial, it seems like there's something weird going on with the ari scores. For both the noiseless and noisy case, the score are exactly the same. Also, for the noisy case, the printout shows MVMDS having a better score, but it says pca performs better.

Fixed. Cells were run out of order.

PR #146, opened 2020-04-27
Implemented Co-Regularized Multi-view Spectral Clustering

### LINKED ISSUE #138: Spectral Clustering n_init parameter is unused
## Expected Behavior
MultiviewSpectralClustering n_init parameter should be used to do multiple random initializations of the k-means clustering part of the algorithm.

## Actual Behavior
The n_init parameter is taken in and used to set self.n_init, but k-means is only run once.

## Template Code
```mvlearn
cluster1 = MultiviewSpectralClustering(n_init=1)
cluster2 = MultiviewSpectralClustering(n_init=100)
```
These perform the exact same every time.

## Your Environment
* Python version: 3.7
* mvlearn version: 0.1.0

alexc3071:
#### Reference Issues/PRs
Fixes #138 

This pull request implements the Co-Regularized Multi-view Spectral Clustering algorithm along with the associated tutorial notebooks and unit tests. This also makes relevant changes to the Multi-view Spectral Clustering code to make it easier to reuse methods and exception-handling functionality for the co-regularized version.


mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,239 @@
...
+    Notes
+    -----
+
+    In standard spectral clustering, the eigenvector matrix U for a given view
+    is the new data representation to be used for the subsequent k-means
+    clustering stage. In this algorithm, the objective function has been
+    altered to encourage the pairwise similarities of examples under the new
+    representation to be similar across all views.
+
+    *** Put the pairwise spectral clustering objective here. ***
```
gmischl1, line 90: Does this still need to be added?

gmischl1:
Add the tutorial notebooks that you want to show on the docs to the docs/tutorials.rst file

gmischl1:
Tutorials:

Note the changes to the tutorial notebook names made in this PR: https://github.com/neurodata/mvlearn/pull/145
- Make the headers of the notebooks simple following those examples. (And maybe copy the title for the MultiviewSpectralClustering tutorial so it doesn't change in this PR).

Is there a reason you always only use the first 2 views of UCI for these tests? Does it take a super long time to run on all 6 views? It might be more interesting and demonstrate how this can be used for more than just 2 views if you did it with all 6. Also, I'm wondering if this will help coregularized to better than normal spectral on UCI, since right now it seems like it does worse based on the two tutorial notebooks.

mvlearn/cluster/mv_spectral.py
```diff
@@ -161,60 +161,13 @@ class MultiviewSpectralClustering(BaseEstimator):
     0.872
 
     '''
-    def __init__(self, n_clusters=2, n_views=2, random_state=None,
+    def __init__(self, n_clusters=2, random_state=None,
```
gmischl1, line 164: Change the docstring Parameters section to reflect this change (no more n_views)

mvlearn/cluster/mv_coreg_spectral.py
alexc3071: Yeah, I think I will add it to the docs.

mvlearn/cluster/mv_spectral.py
alexc3071: Will do

mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,237 @@
...
+
+    n_neighbors : int, optional, default=10
+        Only used if nearest neighbors is selected for affinity. The
+        number of neighbors to use for the nearest neighbors kernel.
+
+    v_lambda : float, optional, default=2
+        The regularization parameter. This parameter trades-off the spectral
+        clustering objectives with the degree of agreement between each pair
+        of views in the new representation. Must be a positive value.
+
```
gmischl1, line 80: Can you reorder these parameters so their order matches up with the order they are in the __init__() function?

mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,237 @@
...
+        V_mat = np.hstack(U_mats)
+        norm_v = np.sqrt(np.diag(V_mat @ V_mat.T))
+        norm_v[norm_v == 0] = 1
+        self._embedding = np.linalg.inv(np.diag(norm_v)) @ V_mat
+
+        # Perform kmeans clustering with embedding
+        kmeans = KMeans(n_clusters=self.n_clusters, n_init=self.n_init,
+                        random_state=self.random_state)
+        predictions = kmeans.fit_predict(V_mat)
+        return predictions
```
gmischl1, line 237: Does this method not have a way to do .fit()? If not, can you add something, maybe to the Notes section, saying it doesn't implement fit(), only fit_transform together? See https://github.com/neurodata/mvlearn/blob/master/mvlearn/embed/mvmds.py for an example. Also, can you add this same thing to MultiviewSpectralClustering for consistency?

Let me know what you think about this.

mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,237 @@
...
+    representation to be similar across all views.
+
+    The modified spectral clustering objective for the case of two views is
+    shown and derived in [#3Clu]. In the clustering objective, the
+    hyperparameter lambda trades-off the spectral clustering objectives and
+    the disagreement term.
+
+    For a fixed lambda and n, the objective function is bounded from above and
+    non-decreasing. As such, the algorithm is guaranteed to converge.
+
```
gmischl1, line 97: Can you also add a simple Examples section showing like a 5-10 line example of using this? You'll need to set the seed of numpy if you're using random data, and definitely the random_state of your cluster object to get it to pass the doctests

gmischl1:
Can you also change the multiview_vs_singleview_clustering tutorial simply by removing the n_views=6 wherever it has that?

mvlearn/cluster/mv_coreg_spectral.py
achang63: Ok, I decided follow the convention used by sklearn's spectral clustering algorithm which has both fit and fit_predict. The only functional difference between the two is that one returns the labels at the end and the other doesn't.

mvlearn/cluster/mv_coreg_spectral.py
achang63: Will do

mvlearn/cluster/mv_coreg_spectral.py
achang63: Ok, will do

gmischl1:
In all the tutorial or validation notebooks, don't use reference formatting like [#3Clu] becuase they don't render in the notebooks. Probably just put the authors and title of the paper, maybe the year. Just be consistent. Otherwise, looks good!

mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,290 @@
...
+        Parameters
+        ----------
+        X : array-like, shape(n_samples, n_samples)
+            The similarity matrix for the data in a single view.
+
+        Returns
+        -------
+        la_eigs : array-like, shape(n_samples, n_clusters)
+            The top n_cluster eigenvectors of the normalized graph
```
gmischl1, line 160: it seems like this function actually returns a 3-item tuple, right?

mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,290 @@
...
+                l_comp = np.zeros((n_items, n_items))
+                for v2 in range(self._n_views):
+                    if v1 != v2:
+                        l_comp = l_comp + U_mats[v2] @ U_mats[v2].T
+                l_comp = (l_comp + l_comp.T) / 2
+                # Adding the symmetrized graph laplacian for view v1
+                l_mat = L_mats[v1] + self.v_lambda * l_comp
+                U_mats[v1], d_mat, _ = sp.sparse.linalg.svds(l_mat,
+                                                             k=self.n_clusters)
+                obj_vals[0, it] = np.sum(d_mat)
```
gmischl1, line 238: is this correct? why is this being updated in the inner loop if the indices [0,it] only depend on the outer loop and the d_mat changes every time? Should it be [index_of_view, it] instead?

mvlearn/cluster/mv_coreg_spectral.py
```diff
@@ -0,0 +1,290 @@
...
+            # Update U and the objective function value for view 1
+            l_comp = np.zeros((n_items, n_items))
+            for vi in range(self._n_views):
+                if vi != 0:
+                    l_comp = l_comp + U_mats[vi] @ U_mats[vi].T
+            l_comp = (l_comp + l_comp.T) / 2
+            l_mat = L_mats[0] + self.v_lambda * l_comp
+            U_mats[0], d_mat, _ = sp.sparse.linalg.svds(l_mat,
+                                                        k=self.n_clusters)
+            obj_vals[0, it] = np.sum(d_mat)
```
gmischl1, line 249: similar comment as above. Why is the first index 0?

mvlearn/cluster/mv_coreg_spectral.py
achang63: Oh, I forgot to change the documentation. I will take care of this.

mvlearn/cluster/mv_coreg_spectral.py
achang63: The first index should be 0 since we are computing the objective function value for view 1 here (which corresponds to array index 0). In the previously mentioned block of code, it the objective function is being computed for views 2 through n_views.

mvlearn/cluster/mv_coreg_spectral.py
achang63: You're right. It should be [v1, it], and that's what it previously was based on my validation notebooks. I think I had deleted it at some point and then copied and pasted the line from line 249 to add it back but forgot to change the indices.

mvlearn/cluster/mv_spectral.py
```diff
@@ -79,6 +76,15 @@ class MultiviewSpectralClustering(BaseEstimator):
         Only used if nearest neighbors is selected for affinity. The
         number of neighbors to use for the nearest neighbors kernel.
 
+    Attributes
+    ----------
+    labels_ : array-like, shape (n_samples)
```
gmischl1, line 81: can you add a comma to make this "(n_samples,)"

achang63: Will do

PR #150, opened 2020-04-29
MVMDS Distance Addition and Test Edits

akoul1:
This PR adds the ability to input distance matrices rather than raw data to MVMDS so that it can be used with any distance metric the user desires.


mvlearn/embed/mvmds.py
```diff
@@ -41,6 +41,9 @@ class MVMDS(BaseEmbed):
 
     num_iter: int (positive), default=15
         Number of iterations stepwise estimation goes through.
+        
+    distance: bool, default=False
```
bdpedigo, line 45: look at how sklearn handles this - they usually have an attribute called `dissimilarity` or `metric`, and you can input things like "euclidean" or "precomputed". Where "precomputed" is what you describe here.

bdpedigo: i would try to follow their API as much as possible

gmischl1: yeah definitely try to match sklearn for this kind of thing

gmischl1:
Note, the branch has a conflict because recently, several algorithms were changed so that attributes which are shown to the user (components in this case) were given a trailing underscore. So, components became components_. Make sure this is changed everywhere in your code, and also in the tests

PR #160, opened 2020-05-03
Shuffle test

### LINKED ISSUE #152: Gaussian Mixture Test Shuffle failing
## Expected Behavior
test_shuffle in test_gaussian_mixture should work every time per correct seeding/implementation.

## Actual Behavior
image

gmischl1:
#### Reference Issues/PRs
Fixes #152 

This changes the test for gaussian mixture shuffling by greatly increasing the number of samples generated before the shuffle. Previously, there were only 10 samples being generated for each dataset, and then they were both shuffled, before comparing to make sure the labels were not in the same order. However, the labels are binary, so by non-zero bad luck, both datasets could be rearranged in an order such that the labels looked the same.


rflperry:
you can just set a seed to guarantee performance every time, even with low samples. I think that's preferable, as technically this can still fail (rarely)

PR #163, opened 2020-05-06
Changelog

### LINKED ISSUE #162: Add a Changelog
Add a changelog to the online documentation so that users can see what is updated.

A simple example from another package: https://pyts.readthedocs.io/en/stable/changelog.html

gmischl1:
#### Reference Issues/PRs

Fixes #162 

This adds a Changelog to the documentation. It includes 0.1.0 as well as the changes we will be adding in 0.2.0.

It also fixes a minor grammar issue in ctclassifier docs.

#### Any other comments?
This PR shouldn't be merged probably until right before we are going to release 0.2.0, so we can make any last minute changes to the changelog if needed. Just putting this PR up to prepare.


gmischl1:
> where did the joblib requirement come from?

It will come from Pr #164

docs/install.rst
```diff
@@ -99,13 +99,14 @@ Python package dependencies
 - scikit-learn >=0.19.1
 - scipy >=1.1.0
 - seaborn >=0.9.0
+- joblib >=0.11
```
rflperry, line 102: I don't see this in the requirements.txt file? Should be consistent between the two.

gmischl1: It's in the requirements.txt file for the PR #164, which I assumed would be merged first

PR #164, opened 2020-05-07
Multi-view KMeans Update

achang63:

This pull request makes some minor additions to the multi-view kmeans and spherical kmeans input parameter set and makes small alterations to variable names to make them consistent with those of the other clustering algorithms. In addition, this PR addresses an issue with the multi-view spherical kmeans clustering tutorial not running properly due to a missing spherecluster dependency.


mvlearn/cluster/mv_kmeans.py
```diff
@@ -15,6 +15,7 @@
 # Implements multi-view kmeans clustering algorithm for data with 2-views.
 
 import numpy as np
+from joblib import Parallel, delayed
```
gmischl1, line 18: I think we don't currently have a dependency on joblib. I guess we could add one, since scikit-learn already depends on them, but we'll have to add them to the requirements.txt

gmischl1:
In the validation notebook, it looks like using "[#2Clu]" doesn't render as a reference. I'd probably just put the name of the paper

mvlearn/cluster/mv_kmeans.py
gmischl1: If you can add joblib>=0.11 to requirements.txt, then this should be good to go

gmischl1: That's just the joblib version that sklearn requires. If you need a different version, you can do another. But if 0.11 is fine I'd say do that

achang63: Ok, I'll do that.

mvlearn/cluster/mv_kmeans.py
```diff
@@ -453,66 +532,36 @@ def fit(self, Xs):
...
-            max_iter = self.max_iter
 
         # Type and value checking for n_init parameter
         if not (isinstance(self.n_init, int) and (self.n_init > 0)):
             msg = 'n_init must be a nonnegative integer'
             raise ValueError(msg)
 
+        # Type and value checking for tol parameter
+        if not (isinstance(self.tol, float) and (self.tol >= 0)):
+            msg = 'n_init must be a nonnegative integer'
```
gmischl1, line 547: I think this should say "tol must be a nonnegative float"

mvlearn/cluster/mv_kmeans.py
```diff
@@ -154,7 +164,8 @@ class MultiviewKMeans(BaseKMeans):
     '''
 
     def __init__(self, n_clusters=2, random_state=None, init='k-means++',
-                 patience=5, max_iter=None, n_init=5):
```
gmischl1, line 157: can you reorder the parameters, either here or in the docstrings, so they're the same order?

gmischl1:
Do you need to change add any tests because of these changes?

mvlearn/cluster/mv_kmeans.py
achang63: You're right

achang63:
> Do you need to add any tests because of these changes?

Yes. I'll add the appropriate unit tests.

mvlearn/cluster/mv_kmeans.py
achang63: Will do

PR #165, opened 2020-05-07
Gaussians shuffle

### LINKED ISSUE #144: Gaussian Mixture seed sets both dimensions
## Expected Behavior
Setting a seed should allow for replication of the additive noise but should still allow the noise dimensions to be independent.

## Actual Behavior
The seed sets both noise additions, so the noise for each view is the same (dependent)

gmischl1:
#### Reference Issues/PRs

Fixes #144 

Ensures that the views have reproducible noise dimensions added, but that they are different from each other.


gmischl1:
closing because wrong branch to PR from

PR #173, opened 2020-05-18
UCI View Selection

### LINKED ISSUE #143: Add UCI view selection
## Desired Behavior
load_UCImultifeatures should be able to return a subset of the multiview data

`Xs, y = load_UCImultifeatures(views=[0,1])` should return an `Xs` of the first two views, i.e. `len(Xs) = 2`.

achang63:
#### Reference Issues/PRs

Fixes #143 

This PR implements an option to select which views of the UCI dataset to return. This also removes some redundant label processing.

### DIFF
--- a/mvlearn/datasets/base.py
+++ b/mvlearn/datasets/base.py
@@ -16,7 +16,7 @@ from os.path import dirname, join
 import numpy as np
 
 
-def load_UCImultifeature(select_labeled="all", shuffle=False,
+def load_UCImultifeature(select_labeled="all", views="all", shuffle=False,
                          random_state=None):
     r"""
     Load the UCI multiple features dataset [#1Data]_, taken from the UCI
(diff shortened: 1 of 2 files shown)


mvlearn/datasets/base.py
```diff
@@ -90,6 +98,14 @@ def load_UCImultifeature(select_labeled="all", shuffle=False,
         raise ValueError("If selecting examples by label, must select "
                          "at least 1 and no more than 10.")
 
+    views = list(dict.fromkeys(views))
+    if len(views) == 0:
+        raise ValueError("If views, must select at least 1.")
```
gmischl1, line 103: Just for better wording, I'd say "if choosing specific views, must select at least 1". What do you think?

mvlearn/datasets/base.py
```diff
@@ -103,22 +119,22 @@ def load_UCImultifeature(select_labeled="all", shuffle=False,
...
                 raise ValueError("Bad label: labels must be  in 0, 1, 2,.. 9")
             indices = np.nonzero(labels == label)
             datatemp[j * 200: (j+1) * 200, :] = data[i][indices, :]
-            selected_labels[j*200:(j+1)*200] = labels[indices]
 
         np.random.seed(random_state)
         np.random.shuffle(datatemp)
         selected_data.append(datatemp)
 
+    selected_labels = np.zeros((200*len(select_labeled),))
```
gmischl1, line 135: good call

mvlearn/datasets/base.py
achang63: Yeah, I like that wording too. I'll change it to that.

PR #174, opened 2020-05-19
Train test split

### LINKED ISSUE #116: Wrap train_test split for multi-view data
## Summary of enhancement
For multi-view data, if a user wants to do a train_test_split, like with sklearn, they would need to write a loop and set the random state each time in order to do that to all the views. It might be helpful to wrap sklearn's train_test_split so it works on multi-view data.

For example...

```mvlearn
from mvlearn.utils import train_test_split
Xs_train, Xs_test, y_train, y_test = train_test_split(Xs, y=y, random_state=None)
```
Or
```mvlearn
from mvlearn.utils import train_test_split
Xs_train, Xs_test = train_test_split(Xs, random_state=None)
```

achang63:
#### Reference Issues/PRs

Fixes #116 

This PR implements a utility called train_test_split that wraps the train_test_split function from sklearn.model_selection, which in turn provides an efficient method of obtain training and testing subsets from multi-view data. This PR also adds unit tests.


gmischl1:
My first main comment is that it would be great if it could handle targets (y) as well. We only have one supervised method right now, but it would be helpful to use it with that, and in case we add more in the future. 

I think the tutorial looks fine. It's nice and simple, though maybe showing an example with targets would also be good.

rflperry:
Can we just name it `train_test_split`? It should generalize the sklearn version so this is fine. Idk if we need a tutorial for just this, it's fairly simple. I think this should end up in tutorial in combinations with other functions. The tutorial now is basically what the docstring example should be, though it would be nice if the print was more horizontal, just from a space consideration.

achang63:
> Can we just name it `train_test_split`? It should generalize the sklearn version so this is fine. Idk if we need a tutorial for just this, it's fairly simple. I think this should end up in tutorial in combinations with other functions. The tutorial now is basically what the docstring example should be, though it would be nice if the print was more horizontal, just from a space consideration.

Ok, I'll just turn the tutorial into a docstring example and remove the notebook for the tutorial.

achang63:
> My first main comment is that it would be great if it could handle targets (y) as well. We only have one supervised method right now, but it would be helpful to use it with that, and in case we add more in the future.
> 
> I think for the tutorial, other than also showing targets once that is added, it would be good to try to show the 3 views more distinctly. Maybe creating the 3 views separately and printing them out along with some text saying "view 1: " ... "view 2: " ... Also, I might make the views just a couple samples bigger, like maybe at least 5 each to make it more clear.

Ok, I'll add an option to handle targets (y) as well. As for the tutorial, I'll try if I can use 5 examples for each view without the output seeming too cluttered. However, since I'm turning it into a docstring example anyway, I think that keeping the examples really small might make it more readable.

gmischl1:
> > My first main comment is that it would be great if it could handle targets (y) as well. We only have one supervised method right now, but it would be helpful to use it with that, and in case we add more in the future.
> > I think for the tutorial, other than also showing targets once that is added, it would be good to try to show the 3 views more distinctly. Maybe creating the 3 views separately and printing them out along with some text saying "view 1: " ... "view 2: " ... Also, I might make the views just a couple samples bigger, like maybe at least 5 each to make it more clear.
> 
> Ok, I'll add an option to handle targets (y) as well. As for the tutorial, I'll try to see if I can use 5 examples for each view without the output becoming too cluttered. However, since I'm turning it into a docstring example anyway, I think that keeping the examples really small might make it more readable.

Yeah I agree the Examples section should be kept minimal but still informative, so small is fine for that.

mvlearn/preprocessing/preprocessing.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            Xs_test.append(splits[i])
+
+    y_train = y_test = None
+
+    if y is not None:
+        y_train = Xs_train.pop()
+        y_test = Xs_test.pop()
+
+    return Xs_train, Xs_test, y_train, y_test
```
gmischl1, line 153: Doesn't sklearn return a list of length 2*n_inputs? Maybe make it a list, though I see how then it becomes a list of lists of arrays lol

Also, I think rather than returning None for the y_train and y_test, they should only be returned if they are passed in.

What do you think?

rflperry: yeah I wouldn't return none, because then it's a sort of weird 2 input 4 output situation. It's unintuitive definitely.

mvlearn/preprocessing/preprocessing.py
```diff
@@ -0,0 +1,153 @@
...
+                                                  stratify=stratify)
+    else:
+        Xs, y = check_Xs_y(Xs, y)
+        splits = model_selection.train_test_split(*Xs, y, test_size=test_size,
+                                                  train_size=train_size,
+                                                  random_state=random_state,
+                                                  shuffle=shuffle,
+                                                  stratify=stratify)
+
+    Xs_train = list()
```
pierreablin, line 139: rather use Xs_train = []

mvlearn/preprocessing/preprocessing.py
```diff
@@ -0,0 +1,153 @@
...
+    Xs_test = list()
+    for i in range(len(splits)):
```
pierreablin, line 141: Rather use the one liners

Xs_train = splits[::2]
Xs_test = splits[1::2]

mvlearn/preprocessing/preprocessing.py
```diff
@@ -0,0 +1,153 @@
...
+        if i % 2 == 0:
+            Xs_train.append(splits[i])
+        else:
+            Xs_test.append(splits[i])
+
+    y_train = y_test = None
+
+    if y is not None:
+        y_train = Xs_train.pop()
```
pierreablin, line 150: maybe the branching should be done in the previous branching, where there is the `if y is None` branch. IMO it would make things more readable.

mvlearn/preprocessing/preprocess.py
```diff
@@ -0,0 +1,158 @@
...
+# limitations under the License.
+# Authors: Alexander Chang, Gavin Mischler
+
+import numpy as np
+from sklearn import model_selection
+from itertools import chain
+from ..utils.utils import check_Xs_y, check_Xs
+
+
+def train_test_split(Xs, y=None, test_size=None, train_size=None,
```
agramfort, line 22: your function train_test_split is duplicated....

gavinmischler: oops, forgot to remove the old file

mvlearn/preprocessing/preprocess.py
```diff
@@ -0,0 +1,158 @@
...
+    Splits multi-view data into random train and test subsets. This
+    utility wraps the train_test_split function from
+    sklearn.model_selection for ease of use.
+
+    Parameters
+    ----------
+    Xs : list of array-likes or numpy.ndarray
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+            The input data.
```
agramfort, line 34:
```suggestion
        The input data.
```

mvlearn/preprocessing/preprocess.py
```diff
@@ -0,0 +1,158 @@
...
+# Authors: Alexander Chang, Gavin Mischler
+
+import numpy as np
+from sklearn import model_selection
+from itertools import chain
+from ..utils.utils import check_Xs_y, check_Xs
+
+
+def train_test_split(Xs, y=None, test_size=None, train_size=None,
+                     random_state=None, shuffle=True, stratify=None):
```
agramfort, line 23: why don't you stick to the sklearn public API?

https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html

in sklearn you can also split sample_weights, groups etc...

mvlearn/preprocessing/preprocess.py
```diff
@@ -0,0 +1,158 @@
...
+        - Xs_test : list of array-likes, testing data.
+        - y_train : numpy.ndarray, training targets, only if y provided
+        - y_test : numpy.ndarray, testing targets, only if y provided
+
+    Examples
+    --------
+    >>> from mvlearn.preprocessing import train_test_split
+    >>> import numpy as np
+    >>> RANDOM_STATE=10
+    >>> np.random.seed(RANDOM_STATE)
```
agramfort, line 81: don't fix global state but use np.random.RandomState

pierreablin:
I think that this function can be moved to model_selection, to mimic sklearn.

mvlearn/preprocessing/preprocess.py
gavinmischler: ok, i'll make sure it can deal with arbitrary numbers of multiview or single-view arrays

mvlearn/preprocessing/preprocess.py
gavinmischler: not sure why this was even in the example

gavinmischler:
I think the circleci is failing because of some of the tutorial notebooks, which I assume will be resolved by #238

agramfort:
yes. That's why I suggested last week to invest collective time on #238
<https://github.com/mvlearn/mvlearn/pull/238> by pushing directly to my PR.

I see there is a already a conflict in #238
<https://github.com/mvlearn/mvlearn/pull/238> and I have not much time
anymore. Please catch the ball :)

>

rflperry:
Yes I didn't have time last week to work on #238 but am prioritizing it this week.

docs/changelog.rst
```diff
@@ -49,6 +49,8 @@ Updates in this release:
 
 - |Major Feature| Adds an ``model_selection`` module with multiview cross validation. `#234 <https://github.com/mvlearn/mvlearn/pull/234>`_ by `Pierre Ablin`_.
 
+- |Feature| Adds the function ``model_selection.train_test_split`` to wrap that of `sklearn <https://scikit-learn.org/stable/whats_new/v0.23.html>`_ for multiview data or items. `#174 <https://github.com/mvlearn/mvlearn/pull/174>`_ by `Alexander Chang <https://github.com/achang63>`_ and `Gavin Mischler <https://gavinmischler.github.io/>`_.
```
agramfort, line 52:
```suggestion
- |Feature| Adds the function ``model_selection.train_test_split`` to wrap that of `scikit-learn <https://scikit-learn.org/>`_ for multiview data or items. `#174 <https://github.com/mvlearn/mvlearn/pull/174>`_ by `Alexander Chang <https://github.com/achang63>`_ and `Gavin Mischler <https://gavinmischler.github.io/>`_.
```

mvlearn/model_selection/split.py
```diff
@@ -0,0 +1,149 @@
...
+        - Xs_train : list of array-likes, training data.
+        - Xs_test : list of array-likes, testing data.
+        - y_train : numpy.ndarray, training targets, only if y provided
+        - y_test : numpy.ndarray, testing targets, only if y provided
+
+    Examples
+    --------
+    >>> from mvlearn.preprocessing import train_test_split
+    >>> import numpy as np
+    >>> RANDOM_STATE=10
```
agramfort, line 74:
```suggestion
```

mvlearn/model_selection/split.py
```diff
@@ -0,0 +1,149 @@
...
+    >>> print(y_train)
+    [2 1]
+    >>> print(y_test)
+    [0]
+    '''
+
+    splitting = []
+    for a in inputs:
+        splits = None
+        if isinstance(a, list) or (type(a).__module__ == np.__name__ and
```
agramfort, line 129: isinstance(a, np.ndarray) was not enough? why did you need this `__module__`?

mvlearn/model_selection/split.py
```diff
@@ -0,0 +1,149 @@
...
+                                   a.ndim == 3):
+            splits = ms.train_test_split(*a,
+                                         test_size=test_size,
+                                         train_size=train_size,
+                                         random_state=random_state,
+                                         shuffle=shuffle,
+                                         stratify=stratify)
```
agramfort, line 136: use 


```
kwargs = dict(test_size=test_size, train_size=train_size, random_state=random_state, shuffle=shuffle, stratify=stratify)
```

then **kwargs in both brnaches of the if

tests/test_split.py
```diff
@@ -0,0 +1,87 @@
+import pytest
+import numpy as np
+from mvlearn.model_selection import train_test_split
+
+RANDOM_SEED = 10
+np.random.seed(RANDOM_SEED)
```
agramfort, line 6: never use global seed that have effect across tests but use np.random.RandomState

tests/test_split.py
```diff
@@ -0,0 +1,87 @@
...
+
+@pytest.fixture(scope='module')
+def r_data():
+    data = list()
+    for _ in range(3):
+        data.append(np.random.random((20, 10)))
+
+    labels = np.random.random((20,))
```
agramfort, line 14: use an instance of RandomState

tests/test_split.py
```diff
@@ -0,0 +1,87 @@
...
+                                        train_size=None, random_state=RANDOM_SEED,
+                                        shuffle=True, stratify=None)
+    assert len(train_Xs) == 3
+    assert len(test_Xs) == 3
+
+    for X in train_Xs:
+        assert X.shape == (16, 10)
+    for X in test_Xs:
+        assert X.shape == (4, 10)
+
```
agramfort, line 54: run flake8 on this file

agramfort:
what type are you expecting to catch in this if condition? np.ndarray what
else?

>

gavinmischler:
> what type are you expecting to catch in this if condition? np.ndarray what else?
> […](#)

I was just trying to catch 3D data, either a list of arrays or a 3D array, since that will be the form of multiview data, as well as any any arrays/weights which are 3D

agramfort:
`isinstance(a, np.ndarray) and a.ndim == 3` ?

gavinmischler:
> `isinstance(a, np.ndarray) and a.ndim == 3` ?

Oh yeah ok, I changed to that

gavinmischler:
Added the train_test_split function to tutorials where splitting was previously done with looping or explicit separation

rflperry:
#238 has fixed all the circleci issues. This PR should pass once that gets merged.

examples/embed/plot_kcca_tutorial.py
```diff
@@ -22,10 +22,11 @@
 """
 
 
-from scipy import stats
-from mvlearn.plotting import crossviews_plot
 from mvlearn.embed import KCCA
+from mvlearn.model_selection import train_test_split
+from mvlearn.plotting import crossviews_plot
 import numpy as np
+from scipy import stats
```
agramfort, line 29: imports should go from more generic to local. So first standard lib, then numpy/scipy then mvlearn

mvlearn/model_selection/split.py
```diff
@@ -0,0 +1,140 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Authors: Alexander Chang, Gavin Mischler
+
+import numpy as np
+from sklearn import model_selection as ms
+
+
+def train_test_split(*inputs, test_size=None, train_size=None,
```
agramfort, line 20:
```suggestion
def train_test_split(*inputs, *, test_size=None, train_size=None,
```
to force passing param names

agramfort (approved):

mvlearn/model_selection/split.py
gavinmischler: The build failed when I changed to this, saying Invalid Syntax. A quick test in a jupyter notebook makes me think Python doesn't allow you to use * after your first argument is *inputs rather than inputs. So, I changed to **options for all the kwargs to follow sklearn's API completely

Issue #177, opened 2020-05-20
mimic sklearn datasets API with ours


rflperry:
to what extent we can, parameter names etc. Verify this at least.

achang63:
Should we have the load_UCImultifeature function return a dict-like object like the sklearn data_loaders? This will require us to change and rerun a lot of the tutorials and example doc strings, but the resulting dictionary object can provide users with a lot of useful attributes and metadata associated with the dataset.

gmischl1:
Yeah we probably should allow dictionary output (though still with a return_Xs_y argument so you can get just (Xs, y)). I see that sklearn has the dictionary as the default, I guess we could have the return_Xs_y default to True so we don't have to change all the tutorials. @achang63, @rflperry what are your thoughts? Should the dictionary or the Xs, y be the default?

rflperry:
I'm fine with doing `return_Xs_y` as the default so that tutorials etc. don't have to change. 

FYI this issue also corresponded to the gaussian_mixture function which could better be modified to mimic [make blobs](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.make_blobs.html)

achang63:
Ok, I like that idea. I'll make return_Xs_y the default but give the option to the return the dictionary.

PR #187, opened 2020-05-23
Gaussian mixture api enhancement

rflperry:
#### Reference Issues/PRs
#177 

Changes `mvlearn.datasets.gaussian_mixture.GaussianMixture` parameters to better mimic sklearn's datasets, notably the make blobs.

#### Any other comments?
Also updated tutorials that used it to reflect the changes, all that I could tell used it. Also updated some parameter descriptions.

Comment, deleted some listed attributes that were duplicates of parameters. Inputted parameters get stored as `self.name` if they remain unchanged or `self.name_` if they are changed. Only the latter need to be listed under attributes as the former are implicit from the parameters. See sklearn [api](https://scikit-learn.org/stable/glossary.html#glossary) for details.


mvlearn/datasets/gaussian_mixture.py
```diff
@@ -93,101 +95,113 @@ def __init__(self, mu, sigma, n, class_probs=None, random_state=None,
...
         ----------
-        latent : np.ndarray
-            Latent distribution data. latent[:,i] is randomly sampled from
-            a gaussian distribution with mean mu[i] and covariance sigma[i].
-        Xs : list of array-like
+        latent_ : np.ndarray, of shape (n_samples, n_dims)
+            Latent distribution data. latent[i] is randomly sampled from
+            a gaussian distribution with mean centers[i] and covariance
+            covariances[i].
+        y_ : np.ndarray, of shape (n_samples)
```
gmischl1, line 102: This is probably intuitive what y is, but it's the only attribute without a description

mvlearn/datasets/gaussian_mixture.py
```diff
@@ -93,101 +95,113 @@ def __init__(self, mu, sigma, n, class_probs=None, random_state=None,
...
-            msg = "length of mu and sigma must be equal"
+        if self.centers_.ndim == 1:
+            self.centers_ = self.centers_[np.newaxis, :]
+        if self.covariances_.ndim == 2:
+            self.covariances_ = self.covariances_[np.newaxis, :]
+        if not self.centers_.ndim == 2:
+            msg = "centers is of the incorrect shape"
+            raise ValueError(msg)
+        if not self.covariances_.ndim == 3:
+            msg = "covariances if of the incorrect shape"
```
gmischl1, line 154: should be "is" not "if" I think

PR #199, opened 2020-06-03
Base cluster

gmischl1:
Adds a base class for the cluster module from which all cluster objects should inherit.


rflperry:
Went further with some more updates

cluster module updates
- Singular base class `cluster.base.BaseCluster` with implemented `fit_transform`.
- Removed `fit_transform` from functions as they inherit from base_class.
- Methods now all have a `labels_` attribute computed from a `predict` called in `fit`. Some were doing this already, sklearn does this.
- Added a `y=None` to the base class and to the methods, also what sklearn does.

other updates
- moved cluster tests to their own folder
- added `__init__.py` files to the test folders as otherwise you can't test individual folders necessarily.

PR #225, opened 2020-09-10
Add groupPCA and groupICA

pierreablin:
@hugorichard and myself coded this to add groupca and group ICA to `mvlearn.decomposition`.

**GroupPCA:**
- Take the datasets [X1, ..., Xs], perform an individual pca on each dataset, then stack them in the features direction. Perform a final PCA on the stacked data.

**GroupICA:**
- After GroupPCA, apply a usual ICA algorithm on top, in order to obtain a source matrix S. The estimated individual mixing matrices Ai are then estimated by least-squares estimation, by minimizing |Xi - AiS|.

It is still ungoing work, but I think the API is usable, and the tests are strong enough.

**Still missing:**

- Better references, especially for group pca
- we should agree on the api :)


mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+                )
+            ]
+        else:
+            X = self.grouppca_.transform(Xs)
+            return np.dot(X, self.components_.T)
+
+    def inverse_transform(self, X_transformed):
+        r"""
+        A method to recover multiview data from transformed data
+        """
```
gmischl1, line 281: Unless this is only supposed by be used privately (in which case add an underscore), it should have Parameters and Returns docs

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,293 @@
+import numpy as np
```
gmischl1, line 1: Add copyright and author info to the top of the file

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+
+from ..utils.utils import check_Xs
+from .base import BaseEstimator
+from .grouppca import GroupPCA
+
+
+class GroupICA(BaseEstimator):
+    r"""
+    Group Independent component analysis.
+    As an optional preprocessing, each dataset in `Xs` is reduced with
```
gmischl1, line 30: To me, "optional" is confusing here, because it seems to imply that this preprocessing is optional within this algorithm. It's not though, right? GroupICA necessarily includes GroupPCA? I think it would be more clear to say something like "As a preprocessing step before Independent Component Analysis, each dataset in `Xs`... "
Thoughts?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+    n_subjects_ : int
+        Number of subjects in the training data
+
+    References
+    ----------
+    .. [#1groupica] Calhoun, Vince, et al. "A method for making group
+                    inferences from functional MRI data using independent
+                    component analysis."
+                    Human brain mapping 14.3 (2001): 140-151.
+    Examples
```
gmischl1, line 115: Maybe add a See Also section which links to multiviewica and grouppca?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+import scipy.linalg as linalg
+from sklearn.decomposition import fastica
+
+
+from ..utils.utils import check_Xs
+from .base import BaseEstimator
+from .grouppca import GroupPCA
+
+
+class GroupICA(BaseEstimator):
```
gmischl1, line 27: is there a reason this is a subclass of BaseEstimator instead of BaseDecomposer?

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,293 @@
...
+from sklearn.decomposition import PCA
+
+
+from ..utils.utils import check_Xs
+from .base import BaseEstimator
+
+
+class GroupPCA(BaseEstimator):
```
gmischl1, line 9: is there a reason this is a subclass of BaseEstimator instead of BaseDecomposer?

gmischl1 (commented):
I added comments mostly about the documentation. Let me know if you disagree with anything or have thoughts on any of my questions.

I think the API for returning multiple outputs through an optional parameter is fine as it is here.

Also, with it implemented here, I assume the plan is to delete GroupICA from mv_ica.py altogether, though it's not part of this PR.

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+        if self.multiple_outputs:
+            return self.transform(Xs)
+        else:
+            return sources
+
+    def fit(self, Xs, y=None):
+        r"""Fit the model with Xs.
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, n_features)
```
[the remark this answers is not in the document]
pierreablin: Yes indeed :+1:

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+            self.individual_mixing_.append(lstq_solution.T)
+        self.n_components_ = gpca.n_components_
+        self.n_features_ = gpca.n_features_
+        self.n_samples_ = gpca.n_samples_
+        self.n_subjects_ = gpca.n_subjects_
+        return sources
+
+    def fit_transform(self, Xs, y=None):
+        """
+        Fit  to the data and transform the data
```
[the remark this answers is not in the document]
pierreablin: :+1:

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,293 @@
...
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+        y : array, shape (n_samples,), optional
+
+        Returns
+        -------
+        X_transformed : array of shape (n_samples, n_components)
```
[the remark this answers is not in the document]
pierreablin: You are correct, I will fix this.

mvlearn/decomposition/groupica.py
pierreablin: This should indeed be clearer. The optional preprocessing step consists in reducing each dataset individually, with an individual PCA. Then, the stacking, PCA and ICA are core to the group ICA and not optional preprocessing. I will make it clearer.

mvlearn/decomposition/groupica.py
pierreablin: It's a mistake on my part, good catch.

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,293 @@
...
+            - Xs shape: (n_views,)
+            - Xs[i] shape: (n_samples, n_features_i)
+        y : array, shape (n_samples,), optional
+
+        Returns
+        -------
+        X_transformed : array of shape (n_samples, n_components)
+            The transformed data
+        """
+        Xs = check_Xs(Xs, copy=self.copy)
```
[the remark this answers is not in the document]
pierreablin: I used sklearn.utils.validation.check_is_fitted, what do you think? I see that it is not what is done in other files of mvlearn.

pierreablin:
I just pushed some code that answers your comments. 

I think this PR allows to implement a proper groupPCA, and then just build groupICA on top of it, so I'd say the goal would be to remove mv_ica.GroupICA altogether, yes.

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+            If `multiple_outputs` is True, it is a list with the estimated
+            individual sources.
+            If `multiple_outputs` is False, it is a single array containing the
+            shared sources.
+        """
+        check_is_fitted(self)
+        Xs = check_Xs(Xs, copy=True)
+
+        if self.multiple_outputs:
+            return [
```
pierreablin, line 286: should the output of a `.transform` be a list of 2 d arrays or a 3d array?

rflperry: We typically do a list of 2d arrays. 1) not guaranteed the shapes across views will allow for a 3d array. 2) the interpretation of different views is more clear.

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Authors: Pierre Ablin, Hugo Richard
+
+import numpy as np
+from picard import picard
+import scipy.linalg as linalg
```
agramfort, line 18:
```suggestion
from scipy import linalg
```

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Authors: Pierre Ablin, Hugo Richard
+
+import numpy as np
+from picard import picard
```
agramfort, line 17: import picard last as it relies on scipy

order of import from top to bottom food chain

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+class GroupICA(BaseDecomposer):
+    r"""
+    Group Independent component analysis.
+    Each dataset in `Xs` is reduced with usual PCA (this step is optional).
+    Then, datasets are concatenated in the features direction, and a PCA is
+    performed on this matrix, yielding a single dataset. ICA is finally
+    performed yielding the output dataset S. The unmixing matrix W
+    corresponding to data X are obtained by solving
+    argmin_{W} ||S - WX||^2.
+
```
agramfort, line 37:
```suggestion
    r"""Group Independent component analysis.

    Each dataset in `Xs` is reduced with usual PCA (this step is optional).
    Then, datasets are concatenated in the features direction, and a PCA is
    performed on this matrix, yielding a single dataset. ICA is finally
    performed yielding the output dataset S. The unmixing matrix W
    corresponding to data X are obtained by solving
    argmin_{W} ||S - WX||^2.

```

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+        The number of individual components to extract as a preprocessing.
+        If None, no preprocessing is applied. If an `int`, each dataset
+        is reduced to this dimension. If a list, the dataset `i` is
+        reduced to the dimension `n_individual_components[i]`.
+        If `'auto'`, set to the minimum between n_components and the
+        smallest number of features in each dataset.
+
+    multiple_outputs : bool, optional
+        If True, the `.transform` method returns one dataset per view.
+        Otherwise, it returns one dataset, of shape (n_samples, n_components)
```
agramfort, line 54: this needs to be discussed as I feel it will be very common. How about when you have a pipeline in mvlearn?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+        Chooses which ICA solver to use. `picard` is generally faster and
+        more reliable, but it requires to be installed.
+
+    ica_kwargs : dict
+        Optional keyword arguments for the ICA solver. If solver='fastica',
+        see the documentation of sklearn.decomposition.fastica.
+        If solver='picard', see the documentation of picard.picard.
+
+    random_state : int, RandomState instance, default=None
+            random state
```
agramfort, line 69:
```suggestion
    random_state : int, RandomState instance, default=None
        Controls the random number generator used in the estimator. Pass an int for reproducible output across multiple function calls.
```

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+
+    n_samples_ : int
+        Number of samples in the training data.
+
+    n_subjects_ : int
+        Number of subjects in the training data
+
+    See also
+    --------
+    grouppca
```
agramfort, line 112: does the cross-refs work in the doc?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+        self.n_individual_components = n_individual_components
+        self.multiple_outputs = multiple_outputs
+        self.prewhiten = prewhiten
+        self.solver = solver
+        self.ica_kwargs = ica_kwargs
+        self.random_state = random_state
+
+    def _fit(self, Xs, y=None):
+        """
+        Fit  to the data and transform the data
```
agramfort, line 165: run pydocstyle on this code

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,326 @@
...
+        X_transformed : list of array-likes or numpy.ndarray
+            The transformed data.
+            If `multiple_outputs` is True, it is a list with the estimated
+            individual sources.
+            If `multiple_outputs` is False, it is a single array containing the
+            shared sources.
+        """
+        sources = self._fit(Xs, y)
+        if self.multiple_outputs:
+            return self.transform(Xs)
```
agramfort, line 241: fit_transform must be the same as fit().transform

agramfort (commented):
here is a first batch of comments

mvlearn/decomposition/groupica.py
pierreablin: Indeed it's an important question. I think that it is ok to have a single dataset as output, because is allows for seamless integration with sklearn. To me, the most important thing is that the user is well aware of this.

mvlearn/decomposition/groupica.py
pierreablin: It is the same so far, the `test_transform` function in `tests/decomposition/test_groupica.py` makes sure that is behaves like this. 

I thought that this was the simplest way to not have duplicate code: the `_fit` function returns the shared sources, which is the expected output of `fit_transform` is `multiple_outputs` is False. Otherwise, we still need to compute the individual sources.

mvlearn/decomposition/groupica.py
agramfort: to integrate with sklearn you can have ConcatTransformer that you insert in a pipeline?

mvlearn/decomposition/groupica.py
pierreablin: Ok, so this code is fine.

mvlearn/decomposition/groupica.py
pierreablin: I don't know, I just see that it is how it is done in the other packages.

mvlearn/decomposition/groupica.py
pierreablin: Yes I think this is something that definitely should be implemented.
I'm not sure it is in the library yet, any idea @rflperry ?

Anyways, if `multiple_outputs=False`, we can already integrate this in a sklearn pipeline:



```
import numpy as np

from sklearn.decomposition import PCA
from sklearn.pipeline import Pipeline
from mvlearn.decomposition import GroupICA


pipe = Pipeline([('groupica', GroupICA()),
                 ('pca', PCA())])

pipe.fit_transform(np.random.randn(2, 10, 3))

```

mvlearn/decomposition/groupica.py
rflperry: It will appear as 
image
but we don't have functioning links.

mvlearn/decomposition/groupica.py
rflperry: We have no pipeline module in the library but I am fully in support of adding it as it seems useful.

I see that `multiple_output=False` is the default. However, there will be decomposition algorithms which potentially can only output multiple outputs and thus `multiple_outputs=True`. I think it would be confusing to have some return multiple, and some return just one by default. My stance is to have `True` as default unless every algorithm can output a single view by default.

pierreablin: - I think that the sklearn pipeline works out of the shelf in mvlearn, as there is no check that the output of a .transform is a 2d array in `sklearn.pipeline.Pipeline`. For instance, 

```
import numpy as np

from sklearn.decomposition import PCA
from sklearn.pipeline import Pipeline
from mvlearn.decomposition import GroupICA


pipe = Pipeline([('groupica1', GroupICA(multiple_outputs=True)),
                          ('groupica2',GroupICA())])

pipe.fit_transform(np.random.randn(2, 10, 3))
```
work perfectly fine. To me, @agramfort 's comment was more about transforming multiview datasets into singleview datasets, so that for instance it is painless to do a ridge regression on the stacked features, just by using a pipeline.
I will give the PR a try.

- I share your concern. I make multiple_output=False the default for both GroupPCA and GroupICA because it is the most natural: both methods work by merging the datasets in a single one. Obtaining multiple outputs requires to perform some weird least-squares estimation, which is not really natural to me. Also, it is not obvious how to recover the singles output from the multiple outputs (e.g. averaging does not work)

It would make things more consistent to always output multiple views by default, but it comes at the price of what is natural for some methods.

agramfort:
I would investigate if you can just use a sklearn pipeline (maybe you can
skip the checks)

and yes I suspected the links don't work in see also due to case
sensitive mistakes

>

mvlearn/decomposition/groupica.py
rflperry: The unnatural least squares thing is the computation of the individual unmixing matrices yes? I do see in the group ICA paper they back-construct individual results as well as the group result so that would be a clear multiview output option.

It seems to me that in the generalized case, a decomposition algorithm will compute some combination of individual and common results. I see two ways to return a result:
1) Return the common decomposition result and individual results are left as attributes.
2) Return a list of individual results (whether combined with the common result or not is algo-specific) and leave the common result as an attribute.

In case 2, there is then the issue if an algorithm only computes a common result. One could just return a list of that output repeated since in a way each view is decomposed to the same thing. Of course, that isn't particularly efficient but is a viable solution.

agramfort:
we need to make a choice and stick to it.

one solution is to do only one solution (easy one and more future proof)
and add the other option when / if you feel the urgent need.

mvlearn/decomposition/groupica.py
rflperry: @agramfort comment below supports option 1 I do believe. I am satisfied with this as the default `mvlearn.decomposition` api although the transform api in the `mvlearn.embed` module (where all the cca's lie) is a list output. Unless one has an issue with this across-module discrepancy I will update our API documentation and this PR can proceed with a single output.

agramfort:
If possible all estimators should have the same API so they can be easily swapped

rflperry:
Prioritizing consistent output of the `.transform()` call would imply always outputting a list or otherwise one would need a new method name for one of the modules, which seems undesirable.

pierreablin:
> I would investigate if you can just use a sklearn pipeline (maybe you can skip the checks) and yes I suspected the links don't work in see also due to case sensitive mistakes
> […](#)

- I did a quick check, it seems that sklearn pipelines already work with mvlearn classes, even if the output is a list of arrays, so that's good news :)


- Regarding the consistent API, on the other hand, I cannot think of a multiview transformer for which is does not really make sense to output one dataset per view (maybe you know one?)

For now on I'll make the `multiple_outputs` to True by default, and add a similar option in grouppca. As said, we can always revert if it's too much of a burden.

agramfort:
I would not expose the multiple_outputs parameter.

you can do

X = np.concatenate(Xs, axis=1)

if you need it.

>

pierreablin:
I think this parameter is critical. The "natural" outputs of groupICA and groupPCA are single datasets. So people familiar with these methods would expect a single dataset as output.

The main problem is that you **cannot** recover this single output from the multiple outputs, so if you don't allow the user to at least set `multiple_output=True` the user will never be able to recover the natural output of groupica :(

hugorichard:
I think the solution to have the natural output of GroupPCA as an attribute is a good one.

pierreablin:
But this makes an attribute that scales with the number of samples, and you cannot get the single dataset output on testing data, which is sometimes needed.

hugorichard:
Good point.

rflperry:
Trying to get around having a parameter that scales with the number of samples seems difficult for all potential algorithms. But it would be worse to have to store all the multiview outputs rather than a group one. You can't just store attributes that don't vary with sample size and use inputted data to do the computation on call?

I don't see how you can't express the single output as long as you have a `multiple_outputs=False` option and store the proper attributes. I think that the multiple parameters output is desired. Although maybe the name should suggest a joint output rather than individual outputs.

pierreablin:
> Trying to get around having a parameter that scales with the number of samples seems difficult for all potential algorithms. But it would be worse to have to store all the multiview outputs rather than a group one. You can't just store attributes that don't vary with sample size and use inputted data to do the computation on call?


Yes, this is exactly what the`.transform` function does right now. 

> I don't see how you can't express the single output as long as you have a `multiple_outputs=False` option and store the proper attributes. I think that the multiple parameters output is desired. Although maybe the name should suggest a joint output rather than individual outputs.

Yes I think that it is the best solution: have the `.transform` return multiple outputs by default, and have an option in the `init` to specify otherwise. I'll make the needed changes.

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,320 @@
...
+        The estimated number of components.
+
+    n_features_ : list of int
+        Number of features in each training dataset.
+
+    n_samples_ : int
+        Number of samples in the training data.
+
+    n_subjects_ : int
+        Number of subjects in the training data
```
agramfort, line 109: n_subjects if MRI jargon. Call it views?

why do you need to store n_samples_ ?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,320 @@
...
+
+    See also
+    --------
+    grouppca
+    multiviewica
```
agramfort, line 114: can you fix the case so the link hopefully work?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,320 @@
...
+            )
+        else:
+            K, W, sources = picard(
+                X_pca.T, **self.ica_kwargs, random_state=self.random_state
+            )
+            sources = sources.T
+        if K is not None:
+            self.components_ = np.dot(W, K)
+        else:
+            self.components_ = W
```
agramfort, line 195: can you add a test to cover this?

mvlearn/decomposition/groupica.py
agramfort: I don't get why you need to branch here. transform should already take multiple_outputs param into account

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,338 @@
...
+        (n_features) where n_features is the number of
+        features in the dataset `i`.
+
+    n_components_ : int
+        The estimated number of components.
+
+    n_features_ : list of int
+        Number of features in each training dataset.
+
+    n_samples_ : int
```
agramfort, line 116: same comments here

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,338 @@
...
+        self.n_components = n_components
+        self.n_individual_components = n_individual_components
+        self.copy = copy
+        self.prewhiten = prewhiten
+        self.whiten = whiten
+        self.random_state = random_state
+
+    def fit_transform(self, Xs, y=None):
+        """
+        Fit  to the data and transform the data.
```
agramfort, line 158:
```suggestion
        """Fit  to the data and transform the data.

```

see pep 257

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,338 @@
...
+    def fit(self, Xs, y=None):
+        r"""Fit the model with Xs.
+
+        Parameters
+        ----------
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
+            - Xs[i] shape: (n_samples, n_features_i)
+        y : None
+            Ignored variable.
```
agramfort, line 241: insert empty line

mvlearn/decomposition/groupica.py
pierreablin: We don't really need to store `n_samples` apart for information purpose, just like in `sklearn.decomposition.PCA`.

I will change the name the n_views_.

mvlearn/decomposition/groupica.py
pierreablin: It's purely in order to avoid doing the same computation twice:

If multiple_outputs = False, you already have computed the fit_transform in the fitting process. If multiple_outputs = True, then you have some more work to do.

agramfort: don't have time to look closer but it's suspicious to me. Also you branch in transform too...

pierreablin: Well I don't know if there is a more readable way to do, but all tests pass, so it has the expected behavior, and I think that this way of doing avoids unnecessary computations.

agramfort:
so be it!

>

mvlearn/decomposition/groupica.py
hugorichard: done

mvlearn/decomposition/groupica.py
hugorichard: done

mvlearn/decomposition/groupica.py
hugorichard: done

mvlearn/decomposition/grouppca.py
hugorichard: done

agramfort:
now the next move is to add these estimators in the API reference page see https://mvlearn.neurodata.io/references/index.html and also have an example built with sphinx gallery that illustrates what this is useful for.

pierreablin:
I have added a tutorial at docs/tutorials/decomposition/group_ica_tutorial.ipynb

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,276 @@
...
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+        y : array, shape (n_samples,), optional
+
+        Returns
+        -------
+        sources : array, shape (n_samples, n_components)
```
rflperry, line 169: The return of `fit` is typically `self` and `sources_` is left as an attribute that can be accessed. Why the deviation here?

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,276 @@
...
+    def transform(self, Xs, y=None):
+        r"""
+        Transform the data Xs into sources.
+
+        Parameters
+        ----------
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
+            - Xs[i] shape: (n_samples, n_features_i)
+        y : array, shape (n_samples,), optional
```
rflperry, line 221: I slightly prefer a description like `y : ignored, included for API compliance.` so there's no confusion

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,276 @@
...
+            If `multiple_outputs` is False, it must be a single
+            array containing shared sources.
+
+        Returns
+        -------
+        Xs : list of arrays
+            The recovered individual datasets.
+        """
+        check_is_fitted(self)
+
```
rflperry, line 264: I would check the input using `check_Xs` from [`mvlearn.utils`](https://github.com/mvlearn/mvlearn/blob/master/mvlearn/utils/utils.py#L19) to verify the input and it also converts a single view array to a list of that array so users can just pass in X which is a lot more user friendly.

mvlearn/decomposition/groupica.py
```diff
@@ -0,0 +1,276 @@
...
+        self.n_components = n_components
+        self.n_individual_components = n_individual_components
+        self.multiple_outputs = multiple_outputs
+        self.prewhiten = prewhiten
+        self.solver = solver
+        self.ica_kwargs = ica_kwargs
+        self.random_state = random_state
+
+    def fit(self, Xs, y=None):
+        r"""Fit  to the data and transform the data.
```
rflperry, line 158: Probably shouldn't say `transform the data` as it's the fit call. also extra space here.

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,358 @@
...
+    ):
+        self.n_components = n_components
+        self.n_individual_components = n_individual_components
+        self.multiple_outputs = multiple_outputs
+        self.prewhiten = prewhiten
+        self.whiten = whiten
+        self.random_state = random_state
+
+    def fit(self, Xs, y=None):
+        """Fit  to the data and transform the data.
```
rflperry, line 168: Same comment on the use word transform here as before.

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+        This merges datasets together and reduces the dimensionality.
+
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+        y : array, shape (n_samples,), optional
```
rflperry, line 178: Same comment on the `y` description as in multiviewICA.

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,358 @@
...
+
+        Xs is projected on the principal components learned
+        in the training set.
+
+        Parameters
+        ----------
+        Xs: list of array-likes
+            - Xs shape: (n_views,)
+            - Xs[i] shape: (n_samples, n_features_i)
+        y : array, shape (n_samples,), optional
```
rflperry, line 258: Same comment as previously on the description of `y`.

mvlearn/decomposition/grouppca.py
```diff
@@ -0,0 +1,358 @@
...
+        ----------
+        X_transformed : list of array-likes or numpy.ndarray
+            The dataset corresponding to transformed data
+
+        Returns
+        -------
+        Xs : list of arrays
+            The recovered individual datasets
+        """
+        check_is_fitted(self)
```
rflperry, line 319: I think it makes sense to `check_Xs` the `X_transformed` here.

rflperry:
Can we also delete the groupICA from the `mv_ica.py` file in this PR?

rflperry:
[Tutorial](https://github.com/pierreablin/mvlearn/blob/grouppca/docs/tutorials/decomposition/group_ica_tutorial.ipynb) feedback:
- Can we remove the line `sys.path.append('/home/pierre/work/mvlearn')` as it's not reproducible for others. If this is an install issue you can just `pip3 install -e .` from the `/home/pierre/work/mvlearn` folder to do a local install.
- Code blocks 9,10 checking the estimated mixing matrices are not run in the version pushed to github.

Otherwise looks great!

mvlearn/decomposition/grouppca.py
pierreablin: I added it if multiple_outputs=True. Otherwise, the input X_transformed should be a plain 2d array, so no need to call check_Xs on it right?

pierreablin:
I just pushed some commits addressing your comments, thanks @rflperry !

> Can we also delete the groupICA from the `mv_ica.py` file in this PR?

It would require changing the code of multiviewICA since it is used as init, and so far the API of groupica.GroupICA and mv_ica.GroupICA are quite different, maybe we should make another clean PR?

mvlearn/decomposition/grouppca.py
rflperry: Yes you are right

rflperry:
> It would require changing the code of multiviewICA since it is used as init, and so far the API of groupica.GroupICA and mv_ica.GroupICA are quite different, maybe we should make another clean PR?

Okay, I agree with this.

pierreablin:
I pushed one last commit that changes the name of the parameter `multiple_outputs` to `multiview_output`

mvlearn/decomposition/grouppca.py
rflperry:
```suggestion
        """Fit to the data Xs.
```

rflperry (approved):
Added docs reference to GroupPCA. LGTM I'll approve and you can merge when you're ready. Fantastic!

agramfort (approved):
Lgtm 

Merge after updating changelog

pierreablin:
The changelog is updated in #230 , so we can merge here ?

rflperry:
> The changelog is updated in #230 , so we can merge here ?

#230 has drafted other changes. I'll just edit the changelog in this PR specific to this one and then you can merge?

agramfort:
perfect

good to go from my end then !

>

PR #227, opened 2020-09-14
Updated API text

rflperry:
Updates the "API of mvlearn objects" section to reflect recent discussion. Clarifies the default `.transform` output to be multiview and that parameters may allow for single view output. Also clarifies the `.prediction` output to be single view.


rflperry:
@pierreablin does this reflect what we talked about well and our desired transform API?

gmischl1 (approved):

docs/contributing.rst
```diff
@@ -278,40 +278,48 @@ trailing underscore to distinguish them from the constants passes to
...
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Transformers
+^^^^^^^^^^^^
 
-A ``transformer`` object modifies the data it is given. An estimator may
-also be a transformer that learns the transformation parameters. The
-transformer object implements the ``transform`` method, i.e.
+A ``transformer`` object modifies the data it is given and by default outputs
+multiview data, a transformation of each input view. Additional parameters may
+allow for a single view output in the case of some algorithms. An estimator may
```
pierreablin, line 286: maybe we should stick to a fix name. In the groupICA pr, it is `multiple_outputs`, but we can change it.

pierreablin (approved):
It looks good to me :)

docs/contributing.rst
rflperry: Yes, I like defining a specific name. Is `multiview_output` better? Just worried that `multiple` is a bit vague. Could do `unified_output=False` or `single_output`as default. `multi_output` may even be better than `multiple`.

pierreablin: My favorite is  `multiview_output`. I think this is the clearest name.

rflperry: oaky, I'll update the API here and should be reflected in groupICA. Thanks!

rflperry:
@pierreablin Added a section about `merger` objects to reflect the soon-to-merged `merge` module

PR #228, opened 2020-09-14
Add two simple classes to merge multiview datasets into single datasets

pierreablin:

This implements two simple transformers that turn a multiview dataset Xs into a single dataset X:

- StackTransformer performs stacking of the features of each dataset
- MeanTransformer averages the features of each dataset

These transformers can be handy to integrate into sklearn pipelines.


pierreablin:
I am not 100% sure about the name `mvlearn.merge` or the name of the transformer, if you have something in mind that sounds better please say so.

rflperry:
I think the transformer names are just fine.

`merge` feels a bit specific to me. But it sort of reminds me of the `preprocessing` module in sklearn. Although this is actually `postprocessing`. Just `mvlearn.processing` maybe? If it's for pipelines, could make it part of `mvlearn.pipeline`. I can think of other names too.

mvlearn/merge/merge_transformers.py
```diff
@@ -0,0 +1,185 @@
...
+            - Xs[i] shape: (n_samples, n_features_i)
+
+        y
+            Ignored
+
+        Returns
+        -------
+        self : object
+            Transformer instance.
+        """
```
rflperry, line 63: I think a `check_Xs` is needed here.

mvlearn/merge/merge_transformers.py
```diff
@@ -0,0 +1,185 @@
...
+    by averaging across views
+
+
+    Attributes
+    ----------
+    n_feature_ : list of ints
+        The number of feature in each view of the input dataset
+        Must be the same for each dataset.
+
+    n_views_ : int
```
rflperry, line 132: Is never stored during `fit`

rflperry:
This is sort of a conundrum, but these are going to be `.transform` methods that are multiview -> singleview. Doesn't that conflict with the API? It helps that it's in a different module. Maybe we want to use the word `.merge` instead of `.transform` in this case.

pierreablin:
Indeed. I'm not sure how to proceed. One thing I'm sure of is that we should stick to the name `.transform` if we want to be able to integrate it out of the shelf in a sklearn Pipeline.

rflperry:
I think one could view this as a subset of transformers, those which force only a single output. One could even call them Mergers (i.e. StackMerger, MeanMerger) and under the API Mergers implement `.fit`, `.merge`, and have `.transform` as an alias for `.merge` (transform = merge) which makes it compatible with sklearn pipelines.

pierreablin:
I think that I correctly implemented your proposed change. It feels like the right way to proceed, maybe @agramfort  has some feedback about this?

pierreablin:
After discussing with @agramfort , it seems that having a `Merger` class that has a `.merge` function is a bit redundant, and I think that having just a `Merger` class with a usual `.transform` function is simpler. The fact that it is a `Merger` already gives the user knowledge that .transform should return a single dataset.

mvlearn/merge/merge_transformers.py
```diff
@@ -0,0 +1,283 @@
...
+        y : array, shape (n_samples,), optional
+
+        Returns
+        -------
+        X_transformed : numpy.ndarray of shape (n_samples, n_features)
+             The singleview output
+        """
+        pass  # pragma: no cover
+
+    @abstractmethod
```
rflperry, line 78: Don't think this should be abstract as it can function as is for inheriting classes.

mvlearn/merge/merge_transformers.py
```diff
@@ -0,0 +1,283 @@
...
+        self.n_total_features_ = sum(self.n_features_)
+        self.n_views_ = len(self.n_features_)
+        return self
+
+    def transform(self, Xs, y=None):
+        r"""Merge the data by stacking its features.
+
+        The multiple views are transformed into a single view dataset by
+        stacking (i.e. concatenating) the features.
+        Parameters
```
rflperry, line 166: Add newline before

mvlearn/merge/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .merge_transformers import StackMerger, MeanMerger
+
+__all__ = ["StackMerger", "MeanMerger"]
```
agramfort, line 3:
```suggestion
__all__ = ["ConcatMerger", "AverageMerger"]
```

?

rflperry:
> After discussing with @agramfort , it seems that having a `Merger` class that has a `.merge` function is a bit redundant, and I think that having just a `Merger` class with a usual `.transform` function is simpler. The fact that it is a `Merger` already gives the user knowledge that .transform should return a single dataset.

Okay great. Couple small edits to address and then it looks good. I think `ConcatMerger` is correct and `AverageMerger` I am fine with.

pierreablin:
Done :)

agramfort (commented):
LGTM 

now we need it to be added to reference page and with an entry in changelog

rflperry:
Added changelog and reference page which all looks good as compiled html. I just `make html` locally to check, don't know of a better solution. Will approve once travis passes.

pierreablin:
I think that we should ultimately move to sphinx gallery for automatic website generation. We should also setup a CI so that the website is updated automatically whenever there is a commit on master.

rflperry:
> I think that we should ultimately move to sphinx gallery for automatic website generation. We should also setup a CI so that the website is updated automatically whenever there is a commit on master.

As is, it currently uses Sphinx for generation, but only the most recent version is generated to [https://mvlearn.neurodata.io/](https://mvlearn.neurodata.io/). I can switch it to master, we just had debated the issue of having features on the website but not pip installable.

rflperry (approved):

agramfort:
+1 to use sphinx-gallery, remove ipynb from repo and use CIs to keep the
website up to date automatically

>

rflperry:
#231 issue made.

PR #229, opened 2020-09-15
Add a meta transformer to transform each dataset individually

pierreablin:

This PR implements mvlearn.filter.Filter, which allows to easily fit and transform individual sklearn transformers.
For instance if you want to do an individual PCA of each view, you can do:

```
from mvlearn.datasets import load_UCImultifeature
from mvlearn.filter import Filter
from sklearn.decomposition import PCA
Xs, _ = load_UCImultifeature()
filt = Filter(PCA(n_components=2))
X_transformed = filt.fit_transform(Xs)
```
Filter can either take a single sklearn transformer, or a list of individual transformers.

Once again I am not 100% sure about the name of this class/ folder, so any idea is welcome.


gmischl1:
Just a thought, but what do people think about the module name "repeat" with class "RepeatTransform"?

We had started making a meta "preprocessing" module to allow for multiview train-test-split ( PR #174 ), but it never got finished. If the purpose of this module is to provide general meta-operations for multiview data, maybe that function could eventually be added to this same module, so a name which is general enough for both a meta transformer operation and meta preprocessing as well as future meta operations would be good. However, if there are other operations we might want to add which are more specific to either preprocessing or filtering, and we don't want to combine them, then maybe a more specific name like filter is better.

pierreablin:
It sounds better indeed. Ok to move this into the preprocessing module, I think that it's where it belongs.

mvlearn/filter/filter.py
```diff
@@ -0,0 +1,201 @@
...
+    >>> print(len(X_transformed))
+    6
+    >>> print(X_transformed[0].shape)
+    (2000, 2)
+    """
+
+    def __init__(self, transformer):
+        self.transformer = transformer
+
+    def prefit(self, Xs, y=None):
```
rflperry, line 64: Keep this public, or make it a hidden method with `_prefit`?

rflperry:
I could see this being used for analyses outside of preprocessing, but I think that most uses would be in preprocessing.

pierreablin:
I changed the names accordingly, and put them in the preprocessing module.

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,201 @@
...
+        view. One clone of the estimator will correspond to each view.
+        Otherwise, it should be a list of estimators, of length the number of
+        views in the multiview dataset.
+
+    Attributes
+    ----------
+    n_views_ : int
+        The number of views in the input dataset
+
+    transformer_list_ : list of objects of length n_views_
```
agramfort, line 42:
```suggestion
    transformers_ : list of objects of length n_views_
```

I would prefer this. it's more consistent with sklearn that uses estimators_

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,201 @@
...
+
+class RepeatTransform(TransformerMixin):
+    r"""Apply a sklearn transformer to each view of a dataset
+
+    Build a transformer from multiview dataset to multiview dataset by
+    using individual scikit-learn transformer on each view.
+
+    Parameters
+    ----------
+    transformer : a sklearn transformer instance, or a list
```
agramfort, line 31: sklearn would call this base_transformer

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,201 @@
...
+        clones of that transformer, otherwise it is a view of self.transformer.
+
+    Examples
+    --------
+    >>> from mvlearn.datasets import load_UCImultifeature
+    >>> from mvlearn.preprocessing import RepeatTransform
+    >>> from sklearn.decomposition import PCA
+    >>> Xs, _ = load_UCImultifeature()
+    >>> repeat = RepeatTransform(PCA(n_components=2))
+    >>> X_transformed = repeat.fit_transform(Xs)
```
agramfort, line 54:
```suggestion
    >>> Xs_transformed = repeat.fit_transform(Xs)
```

?

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,201 @@
...
+        for transformer, X in zip(self.transformer_list_, Xs):
+            transformer.fit(X)
+        return self
+
+    def fit_transform(self, Xs, y=None):
+        r"""Fit and transform each dataset
+
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
```
agramfort, line 125: if ndarray all views have the same shape? this is not clear from the docstring

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,201 @@
...
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+            The data to fit to.
+
+        y : None
+            Ignored variable.
+
+        Returns
+        -------
+        X_transformed : list of array-likes
```
agramfort, line 135:
```suggestion
        Xs_transformed : list of array-likes
```

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,201 @@
...
+# limitations under the License.
+# Authors: Pierre Ablin
+
+from sklearn.base import clone, TransformerMixin
+from sklearn.utils.validation import check_is_fitted
+
+from ..utils import check_Xs
+
+
+class RepeatTransform(TransformerMixin):
```
agramfort, line 23: I would call this

ViewTransformer

in sklearn we have FunctionTransformer etc.

mvlearn/preprocessing/repeat.py
pierreablin: Yes they should. It is a problem that we encounter everywhere I guess. I don't think that an error should be returned if a 3d array is passed to `.fit`, which assumes that all views have the same number of features. What do you think @rflperry ?

pierreablin:
I have addressed most of your comments. I have changed the name of the method to `ViewTransformer` which is indeed more in line with sklearn names.

mvlearn/preprocessing/repeat.py
```diff
@@ -0,0 +1,202 @@
...
+# limitations under the License.
+# Authors: Pierre Ablin
+
+from sklearn.base import clone, TransformerMixin
+from sklearn.utils.validation import check_is_fitted
+
+from ..utils import check_Xs
+
+
+class ViewTransformed(TransformerMixin):
```
agramfort, line 23:
```suggestion
class ViewTransformer(TransformerMixin):
```

pierreablin: I fixed the typo

agramfort:
ok for the code now on my side. Now it needs an entry in the reference page and we should have a what's new page like sklearn to give credits to the contributors.

ok for you @rflperry for the what's new page ?

mvlearn/preprocessing/repeat.py
rflperry: Yes, this the numpy.ndarray correspond to 3d arrays with all views having the same number of features. An error shouldn't be thrown if a 3d array is passed. It's sort of implicit in the second line
```
- Xs length: n_views
```
that a 3D ndarray has all the same features. The goal of those two lines was to clarify the shape of multiview data

rflperry:
> ok for the code now on my side. Now it needs an entry in the reference page and we should have a what's new page like sklearn to give credits to the contributors.
> 
> ok for you @rflperry for the what's new page ?

I can add a reference page entry. I'm not particularly concerned about not having a tutorial for this.

Yep! We have a [`changelog.rst`](https://github.com/mvlearn/mvlearn/blob/master/docs/changelog.rst) file which details what is new in each release we have had. I assume this is satisfactory? I don't know how sklearn does it, but we have been updating that just prior to a new release. Maybe each PR should edit it?

rflperry:
I support that. Names from here on out.

agramfort:
I let you proceed with that

>

pierreablin:
This PR is good for me. When it's merged, I'll make a quick tutorial to show how to use Mergers and ViewTransformers.

rflperry:
Added changelog and references. Also small reordering of the two methods for consistency. Will approve once Travis passes. Separate tutorial would be great.

rflperry (approved):

PR #239, opened 2020-09-28
Remove groupICA from decomposition/mvica.py

pierreablin:
### What does this implement/fix? Explain your changes.

This removes altogether the GroupICA class from decomposition/mvlearn, and instead useds groupica.GroupICA.


mvlearn/decomposition/mv_ica.py
```diff
@@ -223,15 +224,18 @@ def fit(self, Xs, y=None):
             if self.init not in ["permica", "groupica"]:
                 raise ValueError("init should either be permica or groupica")
             if self.init == "permica":
-                algo = PermICA
+                ica = PermICA(
```
rflperry, line 227: Wondering if a `n_components=self.n_components` is needed here?

rflperry (approved):
Small comment/question but otherwise LGTM. Maybe a different PR question, but we have both `PermICA` and `mvica` here, do we want to also add the `GroupICA` to this file? You may have noticed but there is a `BaseICA` here. I attempted to model the attributes and outputs as close to `sklearn.fastICA` as I could.

agramfort (approved):
@rflperry I think we can merge. If we can remove and we don't break any test or doc I claim we are safe. I tend to say "if there is no doc it does not exist !".

agramfort:
thx @pierreablin

PR #249, opened 2020-10-15
Added MCCA module

idc9:
This PR incorporates the MCCA module from https://github.com/idc9/mvdr into mvlearn. The new code implements a generalization of the the SUMCORR-AVGVAR extension extension of CCA to multiple views. This generalization includes regularization (for high-dimensional data), PCA-MCCA (so called informative MCCA) and is easily kernelized. All computations solve eigenvector or generalized eigenvector problems. Details can be found in https://github.com/idc9/mvdr/blob/main/docs/mcca_gevp_docs.pdf as well as the references therein. 

The primary functionality is as follows

- The mvlearn.mcca.mcca.MCCA() object. This is a sklearn compatible estimator for fitting any of the (non-kernelized) MCCA methods. Its transform method outputs a single array of size (n_samples, n_components), which is the common normalized scores (see the aforementioned document). Its blocks_ attribute is a list of MCCABlock() objects that each implement their own transform function. This lets the user transform data one view at a time.

- The mvlearn.mcca.mcca.mcca_gevp() function. This computes the MCCA generalized eigenvector problem given a list of data matrices.

- The mvlearn.mcca.mcca.i_mcca() function. This computes informative MCCA (i.e. where we first compute a low rank PCA, then apply MCCA). The notation "informative MCCA" is borrowed from [Nicholas Asendorf's very nice PhD thesis](https://deepblue.lib.umich.edu/handle/2027.42/113419) as well as related literature. 

- the mvlearn.mcca.k_mcca.k_mcca_evp_svd() function. This computes kernel, regularized MCCA given a list of kernel matrices. The computation is by solving an eigenvector problem and using SVD to remove singularities (similar to [(Bach and Jordan, 2002)](https://www.di.ens.fr/~fbach/kernelICA-jmlr.pdf)). There are multiple options in the literature for how to setup kernel regularized MCCA and we implement three of these options.

- the mvlearn.mcca.k_mcca.KMCCA() object. This takes a list of data matrices, computes the corresponding kernels then computes kmcca. The user can choose which kernel they want for each block and kernel matrices are computed using sklearn.metrics.pairwise.pairwise_kernels.

- tests and one example notebook are also included.


rflperry:
Thanks @idc9 ! There are some things I'd like to restructure/change and I'll push the edits directly to this fork, just wanted to run them by you before I begin.

1. I'd like to move the code to the `embed` module. This will eventually replace the 2view kcca code.
2. consolidate code into two files in the embed module: `k_mcca` and `mcca`
3. Replace the use of the worse `block` with `view` for consistency
4. turn the jupyter notebook into the equivalent python file (sphinx gallery renders it into the website)
5. make the docstrings more consistent with other files of our
6. This should only implement `transform`, not `predict` in my mind. `inverse_transform` should take in `Xs` as an input and so I believe function like your current `predict`

idc9:
Those all make sense to me!

rflperry:
Bulk of stuff done. Todos, coming soon

- [ ] Remove saved `scores_` attributes as attributes that scale with n are to be avoided. 
- [ ] Add multiview parameter to `__init__` which enables common score output rather than a list
- [ ] For KCCA make `fit_transform` efficiently do both `fit` and `transform`, and not store `Xs_`.
- [ ] Update changelog

rflperry:
@idc9 Let me know what you think. couple points

-  scores aren't stored as attributes (we don't like attributes that scale with `n_samples` but rather found just through `.transform`
- A `multiview_output` parameter in the constructor toggles if transform outputs the common scores (normalized) or a list of scores (as with other functions in the package)
- I removed `precomp_svd` inputs for now. that can be dealt with in a PR update if need and is in the github history

idc9:
That all looks good to me!

examples/embed/plot_mcca_tutorial.py
```diff
@@ -0,0 +1,95 @@
...
+                scatter_kwargs={'alpha': 0.4, 's': 2.0})
+
+###############################################################################
+# Regularized MCCA
+# ^^^^^^^^^^^^^^^^
+#
+# We can add regularization with the `regs` argument
+# to handle high-dimensional data.
+
+# regularization value of .1 for each view
```
gavinmischler, line 54: this doesn't do regularization value of 0.5?

gavinmischler (approved):
Minor documentation stuff but looks good

mvlearn/embed/mcca.py
```diff
@@ -0,0 +1,590 @@
...
+        ----------
+        scores: array-like, shape (n_samples, n_components)
+            The CCA scores.
+
+        Returns
+        -------
+        Xs_hat : list of array-likes or numpy.ndarray
+             - Xs_hat length: n_views
+             - Xs_hat[i] shape: (n_samples, n_features_i)
+            The reconstructed original data
```
gavinmischler, line 172: this description is actually also not rendered well. It might be better to have it as

Xs_hat : list of array-likes or numpy.ndarray
             The reconstructed original data
             - Xs_hat length: n_views
             - Xs_hat[i] shape: (n_samples, n_features_i)

gavinmischler (commented):
Another issue with rendering

mvlearn/embed/base.py
```diff
@@ -68,13 +67,248 @@ def fit_transform(self, Xs, y=None):
...
+    def fit(self, Xs, y=None):
+        r"""
+        Learns decompositions of the views.
+
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+            The data to fit to.
```
gavinmischler, line 97: This one also doesn't render well for some reason. The second bullet point has an extra indent in rendering. Maybe for all things like this we should put the line like "The data to fit to" above the bullet points

mvlearn/embed/mcca.py
gavinmischler: or put an extra line between bullet points and "The reconstructed original data", like https://github.com/mvlearn/mvlearn/blob/master/mvlearn/cluster/mv_spherical_kmeans.py does in fit() for example

PR #250, opened 2020-10-16
Examples cleanup

gavinmischler:
In general, the example notebooks do not all have things like simple explanations at the top, consistent title format, etc, and there were a few broken things like links and math not rendering correctly. This PR fixes many of those little inconsistencies.

#### Any other comments?
I noticed that the example files that don't start with plot_ aren't run by sphinx, but they are still in the gallery. Many of these are the validation scripts proving the algorithms work or using them on various synthetic data, but which we were not including in our old documentation. Should we remove them entirely from the examples? (we can do so by editing the conf.py file https://sphinx-gallery.github.io/stable/configuration.html#parsing-and-executing-examples-via-matching-patterns) It seems a little weird to me to include them but not have them run or show their outputs rendered. I think we should choose one way or the other.

The SplitAE tutorial is one that doesn't start with plot_ but isn't a validation notebook. I assume this was done because it couldn't run on sphinx because of the multi-core requirements or something. Should we leave this one as it is, in the examples gallery but without any rendered outputs?


agramfort:
@gavinmischler yes don't use examples as unit test. Either remove them or maybe tests from them (provided it's not tested already)

examples/cluster/plot_mv_spectral_validation_complex.py
```diff
@@ -1,11 +1,18 @@
 """
-========================================================================\
+=========================================================\
 ===================
-Assessing the Conditionally independence views requirement of Multiview \
+Impact of Conditional Independence of Views on Multiview \
 Spectral Clustering
```
agramfort, line 5: you should rather try to find a title that is less than 80c

agramfort (commented):
can you make sure flake8 is now run on the examples and share the html of the rendered doc on circleci?

gavinmischler:
> can you make sure flake8 is now run on the examples and share the html of the rendered doc on circleci?

@agramfort I added "flake8 examples" to the .travis.yml, but I don't have much circleci experience and I'm not sure what you mean by "share the html of the rendered doc on circleci"

agramfort:
Screenshot 2020-10-16 at 11 30 00

follow this link to see the html @gavinmischler

agramfort (approved):
https://89-206838300-gh.circle-artifacts.com/0/dev/auto_examples/index.html looks good !

rflperry:
@gavinmischler Files that took a while to run such as the SplitAE weren't given a `plot_` prefix. Or, one of the spherical cluster examples used `spherecluster` which isn't in our package and the mvkmeans on the newsgroup dataset which just took a while to run. I don't see a great reason to remove those, as I think they have already been done and bring value to the documentation personally.

agramfort:
how do we guarantee this files are not broken? I anticipate this code with
rot otherwise.

>

rflperry:
I take your point. lgtm

PR #257, opened 2020-11-04
CCA UCI data example

rflperry:
Adds an example showing CCA being used for cross-view information retrieval on the UCI dataset. We didn't have a non-simulated data example for CCA methods.

### DIFF
--- a/examples/embed/plot_cca_retrieval.py
+++ b/examples/embed/plot_cca_retrieval.py
@@ -1,0 +1,72 @@
+"""
+=========================================
+Cross-view information retrieval with CCA
+=========================================
+
+MCCA and KMCCA are used to learn informative relationships
+between 2 views of digits in the UCI multi feature dataset: 76 Fourier
+coefficients of the digit shapes and 2x3 window pixel averages.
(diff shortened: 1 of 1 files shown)


examples/embed/plot_cca_retrieval.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+###############################################################################
+# Load and fit the UCI data
+# -------------------------
+#
+# The CCAs learn transformations of the Fourier view and Pixel view that
+# maximizes the cross-view correlation between samples. Here, we have a Fourier
+# sample X1 with unknown pixels. We transform X1 based on the fitted model
+# and find the closest transformed X2 among fitted pixels that in order
```
gavinmischler, line 30: extra word? should it be "fitted pixels in order to..."

gavinmischler (commented):
Is that really a 3 that the KMCCA found? To me at least it doesn't look like one, so maybe print out the corresponding label from the dataset to show that it is

rflperry:
Very strange, it doesn't give me that result locally. There shouldn't be any randomness in the script, and it gives me consistent results locally.

Issue #252, opened 2020-10-26
Difference between `mvlearn.compose` and ` mvlearn.construct`


pierreablin:
The `mvlearn.construct` contains functions to make a single view dataset into a multiview dataset. In `mvlearn.compose` there are also the `Splitter` methods which do the same thing, I'm wondering if we should merge these two modules?

PR #258, opened 2020-11-05
Merges construct into compose

rflperry:
#### Reference Issues/PRs
Solved #252 

Merges all functionality from `mvlearn.construct` into `mvlearn.compose` as the compose tools are a generalization of construct tools. Updates docs accordingly and changelog. A requirement was removes that was erroneous as well.


docs/changelog.rst
```diff
@@ -7,7 +7,7 @@ Changelog
 .. role:: raw-latex(raw)
    :format: latex
 
-.. |MajorFeature| replace:: :raw-html:`<span class="badge badge-success">Major Feature</span>` :raw-latex:`{\small\sc [Major Feature]}`
+.. |Major Feature| replace:: :raw-html:`<span class="badge badge-success">Major Feature</span>` :raw-latex:`{\small\sc [Major Feature]}`
```
gavinmischler, line 10: Having it without the space is how scikit-learn does it based on their raw source (https://scikit-learn.org/stable/_sources/whats_new/v0.24.rst.txt) and rendered output (https://scikit-learn.org/stable/whats_new/v0.24.html). However, none of our badges seem to render anyway. Is it because of our sphinx theme?

gavinmischler: Oh from here (https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through)

"
The "replace" directive is used to indicate replacement text for a substitution reference. It may be used within substitution definitions only. For example, this directive can be used to expand abbreviations:

.. |reST| replace:: reStructuredText

Yes, |reST| is a long word, so I can't blame anyone for wanting to
abbreviate it.
As reStructuredText doesn't support nested inline markup, the only way to create a reference with styled text is to use substitutions with the "replace" directive:

I recommend you try |Python|_.

.. |Python| replace:: Python, *the* best language around
.. _Python: http://www.python.org/
"

does that mean we're supposed to call them with an underscore after since we're not simply replacing simple text?

rflperry: looking into it. The `.rst` file is correct once the space is fixed. The html shows the element but it's not visually there. Looking into it.

agramfort:
i am surprised that no example or narrative doc needs to be updated here. It is not documented at all?

rflperry:
We don't have any current examples. I did update the `.rst` files accordingly.

gavinmischler (approved):

PR #260, opened 2020-11-05
MultiViewICA: closer implementation to GroupICA

### LINKED ISSUE #244: Transpose mv_ica and perm_ica umixing matrix
`GroupICA` has unmixing matrices (`self.individual_components_`) such that XW^T recovers the source. `PermICA` and `MultiviewICA` have the unmixing matrix such that XW recovers the source (as I had initially implemented it). `GroupICA` version I believe is correct and the latter two should be switched.

Multiview ICA tutorial should be updated as well.

hugorichard:
This Fixes #244 
It also changes the implementation so that 
- we do not store attributes that depend on the number of samples
- attributes name and behavior are consistent with the ones of GroupICA
- takes multiview ica as a dependency instead of duplicating the code in https://github.com/hugorichard/multiviewica
- it can use any ViewTransformer as preprocesing (not necessarly PCA)


gavinmischler:
If multiviewica is going to be a dependency, then it will need to be added to conda-forge so that mvlearn can still be installed with conda

examples/decomposition/plot_mv_ica_tutorial.py
```diff
@@ -22,7 +22,9 @@
 import numpy as np
 import matplotlib.pyplot as plt
 
-from mvlearn.decomposition.mv_ica import MultiviewICA, PermICA, GroupICA
+from mvlearn.decomposition.multiviewica import MultiviewICA
+from mvlearn.decomposition.permica import PermICA
+from mvlearn.decomposition.groupica import GroupICA
```
gavinmischler, line 27: can you change these 3 imports to be from the module, rather than from the file? It just looks simpler/cleaner

mvlearn/decomposition/permica.py
```diff
@@ -0,0 +1,117 @@
...
+    verbose : bool, default=False
+        Print information
+
+    Attributes
+    ----------
+    preproc : ViewTransformer-like instance
+        The fitted instance used for preprocessing
+    W_list : np array of shape (n_groups, n_components, n_components)
+        The unmixing matrices to apply on preprocessed data
+
```
gavinmischler, line 52: For PermICA and MultiviewICA in the docs that are rendered (https://124-206838300-gh.circle-artifacts.com/0/dev/references/decomposition.html#multiview-ica), the inheritted attributes from BaseICA don't seem to show up. Based on this (https://stackoverflow.com/questions/30344289/how-can-i-show-inherited-members-of-a-class-in-my-sphinx-documentation), I think you need to add a line to decomposition.rst for the classes where you want base-class attributes to show up in docs, but not totally sure

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,204 @@
...
+                self.preproc_name = "custom"
+            else:
+                raise ValueError(
+                    "The dimension reduction instance needs"
+                    "a transform and inverse transform method"
+                )
+        self.random_state = random_state
+        self.multiview_output = multiview_output
+
+    def fit(self, X, y=None):
```
gavinmischler, line 101: Should this be Xs, and in the docstring?

mvlearn/decomposition/multiviewica.py
```diff
@@ -0,0 +1,182 @@
...
+            random_state=random_state,
+            verbose=verbose,
+            multiview_output=multiview_output,
+        )
+        self.noise = noise
+        self.max_iter = max_iter
+        self.init = init
+        self.tol = tol
+
+    def fit_(self, Xs, y=None):
```
gavinmischler, line 157: This is just a helper function right, that the user would never call? If so, I think you should remove it from the docs by editing the decomposition.rst (https://github.com/mvlearn/mvlearn/blob/master/docs/references/decomposition.rst) to exclude this, and any other helper functions, from any class where they shouldn't actually be called by the user

gavinmischler (commented):
docs stuff

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,204 @@
...
+        ----------
+        X: list of np arrays of shape (n_voxels, n_samples)
+            Input data: X[i] is the data of subject i
+        """
+        if self.preproc is not None:
+            reduced_X = self.preproc.fit_transform(X)
+        else:
+            reduced_X = X.copy()
+        reduced_X = np.array(reduced_X)
+        unmixings_, S = self.fit_(np.swapaxes(reduced_X, 1, 2))
```
agramfort, line 114: private method start with _ and not end with _ so it should be:
```suggestion
        unmixings_, S = self._fit(np.swapaxes(reduced_X, 1, 2))
```

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,204 @@
...
+        Recover the sources from each view (apply unmixing matrix).
+
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+            - Xs length: n_views
+            - Xs[i] shape: (n_samples, n_features_i)
+            Training data to recover a source and unmixing matrices from.
+
+        y : ignored
```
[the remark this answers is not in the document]
hugorichard: Why is this wrong ? y is indeed not used.

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,204 @@
+from .base import BaseDecomposer
+import numpy as np
+from ..preprocessing.repeat import ViewTransformer
+from sklearn.decomposition import PCA
+from scipy import linalg
```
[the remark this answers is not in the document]
hugorichard: I am not sure what is the correct convention. I put the . and .. imports at the end is that the good practice ?

mvlearn/decomposition/baseica.py
```diff
@@ -167,7 +167,7 @@ def fit(self, X, y=None):
                 self.individual_mixing_.append(lstq_solution.T)
         return self
 
-    def transform(self, X, y):
+    def transform(self, X, y=None):
```
agramfort, line 170: there are no y in transforms in sklearn.

it can be considered a bad decision but it's how it is

mvlearn/decomposition/baseica.py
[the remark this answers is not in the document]
agramfort: i order imports based on hierarchy.

numpy
scipy
scikit-learn

as sklearn depends on scipy which depends on numpy

imports from standard library are always first

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,234 @@
...
+        return self.fit(X).transform(X)
+
+    def inverse_transform(self, S):
+        r"""
+        Transforms the sources back to the mixed data for each view
+        (apply mixing matrix).
+
+        Parameters
+        ----------
+        None
```
agramfort, line 215: S ?

mvlearn/decomposition/baseica.py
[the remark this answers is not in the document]
rflperry: y isn't an input to `transform`. docstring should be deleted

mvlearn/decomposition/permica.py
```diff
@@ -0,0 +1,138 @@
...
+    individual_mixing_ : list of array
+        Individual mixing matrices estimated by least squares.
+        `individual_components_[i]` is an array of shape
+        (n_features, n_components) where n_features is the number of
+        features in the dataset `i`.
+
+    See also
+    --------
+    groupica
+    multiviewica
```
rflperry, line 77:
```suggestion
    GroupICA
    MultiviewICA
```

mvlearn/decomposition/multiviewica.py
```diff
@@ -0,0 +1,203 @@
...
+    individual_mixing_ : list of array
+        Individual mixing matrices estimated by least squares.
+        `individual_components_[i]` is an array of shape
+        (n_features, n_components) where n_features is the number of
+        features in the dataset `i`.
+
+    See also
+    --------
+    groupica
+    permica
```
rflperry, line 116:
```suggestion
    GroupICA
    PermICA
```

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,237 @@
...
+        (apply mixing matrix).
+
+        Parameters
+        ----------
+        X_transformed : list of array-likes or numpy.ndarray
+            The dataset corresponding to transformed data
+
+
+
+
```
rflperry, line 218:
```suggestion
```

rflperry: Empty lines

mvlearn/decomposition/multiviewica.py
rflperry, line 116:
```suggestion
    GroupICA
    PermICA
```

rflperry: Similar correction needed in GroupICA

mvlearn/decomposition/permica.py
rflperry, line 77:
```suggestion
    GroupICA
    MultiviewICA
```

rflperry (commented):
Just some small docs rendering things but otherwise looks good to me

rflperry (approved):

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,233 @@
+# MIT License
+
+# Copyright (c) [2017] [Pierre Ablin and Hugo Richard]
```
agramfort, line 3:
```suggestion
# Copyright (c) [2020] [Pierre Ablin and Hugo Richard]
```

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,233 @@
...
+class BaseICA(BaseDecomposer):
+    """A base class for multiview ICA methods.
+
+    Parameters
+    ----------
+    n_components : int, optional
+        Number of components to extract. If None, n_components is set to
+        the minimum number of features in the dataset.
+
+    preproc: None, 'pca' or a ViewTransformer-like instance,\
```
agramfort, line 40: would a pipeline be easier? rather than making the model accept an estimator as init param?

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,233 @@
...
+        instance, random_state is the random number generator; If
+        None, the random number generator is the RandomState instance
+        used by np.random.
+
+    verbose : bool, default=False
+        Print information
+
+    Attributes
+    ----------
+    preproc_instance : ViewTransformer-like instance
```
agramfort, line 67: only attribute names passed to the init are allowed to end without an underscore

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,233 @@
...
+            if hasattr(preproc, "transform") and hasattr(
+                preproc, "inverse_transform"
+            ):
+                self.preproc_name = "custom"
+                self.preproc_instance = preproc
+            else:
+                raise ValueError(
+                    "The dimension reduction instance needs"
+                    "a transform and inverse transform method"
+                )
```
agramfort, line 123: I don't like this preproc_name and preproc_instance thing.

it we need more attributes than preproc than they should appear in the fit and not the init. They must also end then with underscore

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,233 @@
...
+                [x.dot(w.T) for w, x in zip(self.components_, transformed_X)]
+            )
+        else:
+            return np.mean(
+                [x.dot(w.T) for w, x in zip(self.components_, transformed_X)],
+                axis=0,
+            )
+
+    def fit_transform(self, X, y=None):
+        return self.fit(X).transform(X)
```
agramfort, line 204: if you inherit from base class this should not be necessary

tests/decomposition/test_grouppca.py
```diff
@@ -1,8 +1,28 @@
+# MIT License
+
+# Copyright (c) [2017] [Pierre Ablin and Hugo Richard]
```
agramfort, line 3:
```suggestion
# Copyright (c) [2020] [Pierre Ablin and Hugo Richard]
```

tests/decomposition/test_mvica.py
```diff
@@ -30,15 +30,18 @@
 
 import pytest
 import numpy as np
-from mvlearn.decomposition.mv_ica import _hungarian
+from multiviewica import _hungarian
```
agramfort, line 33: what is multiviewica here ??? relative import with a missing . ?

pierreablin: I think _hungarian is a function in Hugo's repo, but it should be private, so I think it could be avoided here. Or since its a simple function, it could also be copy-pasted.

mvlearn/decomposition/baseica.py
pierreablin: I think multiview ICA should come with a standard groupICA initialization, as it should be the preferred init in 90% of the cases. 

And then if the user wants to do something more complicated he should rather use a pipeline.

mvlearn/decomposition/baseica.py
```diff
@@ -0,0 +1,233 @@
...
+
+import numpy as np
+from scipy import linalg
+from sklearn.decomposition import PCA
+
+from .base import BaseDecomposer
+from ..preprocessing.repeat import ViewTransformer
+
+
+class BaseICA(BaseDecomposer):
```
pierreablin, line 31: Why do we need a BaseICA class? I think that in the short term, we only have GroupICA and multiviewICA implemented in mvlearn, and both work quite differently: groupICA estimates one source matrix from the whole dataset, while multiviewICA outputs multiple sources. 

So I don't think it makes things easier to have a BaseICA class.

pierreablin: Also, if we really need a BaseICA class, then we should make GroupICA inherit from it.

mvlearn/decomposition/permica.py
```diff
@@ -0,0 +1,138 @@
+from .baseica import BaseICA
+from multiviewica import permica
+
+
+class PermICA(BaseICA):
```
pierreablin, line 5: I'm not sure whether this algorithm should be at all in mvlearn. It was just a toy algorithm in our paper, and has never really been studied otherwise. Plus we don't really need it to initialize multiviewICA.

mvlearn/decomposition/baseica.py
hugorichard: Ok I agree let's remove PermICA from mvlearn and merge BaseICA and multiViewICA.

mvlearn/decomposition/permica.py
hugorichard: Ok Let's remove it

tests/decomposition/test_mvica.py
hugorichard: Ok for copy pasting

mvlearn/decomposition/baseica.py
hugorichard: Ok I'll remove the preproc_name we don't really need it

mvlearn/decomposition/baseica.py
hugorichard: The current implementation allows the user to do n_components = 5 and do more or less what Pierre suggests without worrying about anything else.
Preproc is only to be used by advanced users and is indeed equivalent to a pipeline.

docs/references/decomposition.rst
```diff
@@ -7,13 +7,13 @@ Multiview ICA
 -------------
 
 .. autoclass:: MultiviewICA
-    :exclude-members: is_fit, get_params, set_params
+    :exclude-members: is_fit, get_params, set_params, _fit
 
 Permutation ICA
```
gavinmischler, line 12: if permica is being removed from mvlearn, take it out of the rst

tests/decomposition/test_mvica.py
agramfort: yes please no other dependency

mvlearn/decomposition/multiviewica.py
```diff
@@ -0,0 +1,344 @@
...
+
+        Parameters
+        ----------
+        X_transformed : list of array-likes or numpy.ndarray
+            The dataset corresponding to transformed data
+
+
+
+
+
```
agramfort, line 299:
```suggestion

```

mvlearn/decomposition/multiviewica.py
```diff
@@ -0,0 +1,344 @@
...
+        Returns
+        -------
+        Xs_new : numpy.ndarray, shape (n_views, n_samples, n_components)
+            The mixed sources from the single source and per-view unmixings.
+        """
+        if not hasattr(self, "components_"):
+            raise ValueError("The model has not yet been fitted.")
```
agramfort, line 306: you could use the check_is_fitted function:

https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html

mvlearn/decomposition/multiviewica.py
```diff
@@ -0,0 +1,344 @@
...
+        r"""
+        Fits the model to the views Xs.
+
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+            - Xs length: n_views
+            - Xs[i] shape: (n_samples, n_features_i)
+            Training data to recover a source and unmixing matrices from.
+        y : ignored
```
agramfort, line 329: no need to pass y here

requirements.txt
```diff
@@ -7,3 +7,4 @@ scipy>=1.4.0
 seaborn>=0.9.0
 joblib>=0.11
 python-picard >= 0.4
+multiviewica >= 0.0.1
```
agramfort, line 10: remove

tests/decomposition/test_groupica.py
```diff
@@ -1,5 +1,25 @@
-import itertools
-import warnings
+# MIT License
+
+# Copyright (c) [2017] [Pierre Ablin and Hugo Richard]
```
agramfort, line 3:
```suggestion
# Copyright (c) [2020] [Pierre Ablin and Hugo Richard]
```

tests/decomposition/test_mvica.py
```diff
@@ -30,15 +30,22 @@
 
 import pytest
 import numpy as np
-from mvlearn.decomposition.mv_ica import _hungarian
-from mvlearn.decomposition import MultiviewICA, PermICA
+from mvlearn.decomposition import MultiviewICA
+from mvlearn.preprocessing import ViewTransformer
+from sklearn.decomposition import PCA
+import scipy
```
agramfort, line 36: wrong import order

mvlearn/decomposition/multiviewica.py
```diff
@@ -0,0 +1,344 @@
...
+        group ica. Else, use the provided array to initialize.
+
+    preproc: 'pca' or a ViewTransformer-like instance,
+        default='pca'
+        Preprocessing method to use to reduce data.
+        If "pca", performs PCA separately on each view to reduce dimension
+        of each view.
+        Otherwise the dimension reduction is performed using the transform
+        method of the ViewTransformer-like object. This instance also needs
+        an inverse transform method to recover original data from reduced data.
```
agramfort, line 69: what do you think would be a good alternative as pca for a custom ViewTransformer?

tests/decomposition/test_mvica.py
gavinmischler: Is this PR still going to add multiviewica as a dependency? It's in the requirements.txt and is imported in multiviewica.py

hugorichard: Yes I would rather do that for now.

requirements.txt
hugorichard: I would like to keep multiviewica as a dependency for now. At some point, I'll make another pull request to move multiviewICA here but for the time being I think it is simpler if we do not duplicate the implementations.

mvlearn/decomposition/multiviewica.py
hugorichard: SRM for example

hugorichard: Also people might want to use transformers that handle data that cannot hold in memory (this is quite common in fMRI)

pierreablin: But then people can just use a custom preprocessing and use `make_pipeline(preprocessing, multiviewica)`, right?

hugorichard: Yes that is right

requirements.txt
agramfort: a dependency in a software is not cheap. If we use multiviewica I think we should not have an implementation in mvlearn. It should be a soft dependency with a nested import. Let's avoid code duplication

mvlearn/decomposition/multiviewica.py
agramfort: I like this ! it simplies the API

hugorichard:
I don't know what is the correct way to check that multiviewica's dependency is indeed soft. Do we usually check these things ? How should we modify the config file for CircleCI ?

rflperry:
We currently have hard dependencies in `requirements.txt` and a couple soft dependencies which are listed in the `requirements_torch.txt` file as they pertain to pytorch tools. Travis and circleci both have `.yml` files which call `pip install -r` on both, while `setup.py` allows for optional installation of the soft dependencies when installing from pypi.

We don't currently test soft/hard. I can look to also add that, requires new test files, and modification to `travis.yml` runs some tests before installation.

I can make push to this branch with edits to make it an optional dependency, it's a bit tricky. I will either make a single `requirements_extras.txt` file or a `requirements` folder with multiple `.txt` files.

hugorichard:
I think the current behavior is ok. We should just check that tests work when mulltiviewICA is not there.

rflperry:
Ah, just saw your edited comment. I made a commit with those tests, need to fix that commit but is a quick fix. I think it is good as in prior releases have accidentally had that issue.

hugorichard:
Thanks ! It is always best to have more things tested !

rflperry:
Soft dependency installations/tests all set

agramfort:
ok to merge but it's hard to know if our approach with minimal dependencies is working.
We need a CI build that runs with the minimal dependencies

rflperry:
travis/pytest now confirm that all minimal dependency function test run with only minimal dependencies. CI for each would distinguish which examples use which dependencies.

agramfort:
+1 for MRG then

Issue #170, opened 2020-05-14
Add rsm tutorial or demonstration


rflperry:
`construct.random_gaussian_projection` has the sklearn tutorials and documentation but we should add a tutorial notebook for `construct.rsm` demonstrating the efficacy (use the reference paper?) in conjunction with another multiview algorithm.

PR #263, opened 2020-11-12
View construction classes and example

rflperry:
#### Reference Issues/PRs
Addresses #170 

- Switches Random subspace method and random gaussian projection from functions to classes in order to accommodate pipelines and out of sample application.

- Adds an example demonstrating their usage in creating multiview datasets from single view. Runs LDA on the Olivevetti face dataset and shows how in the low sample regime, using multiple LDA classifiers on subsampled features is superior.

- In order make the example, added a `ViewClassifier` object in the `compose` module which wraps a single view classifier, applying it to each view and taking the majority vote as the final output (much like how `ViewTransformer` from the `preprocessing` module wraps a transformer). Indeed this latter similarity makes me consider moving `ViewTransformer` to the `compose` module as a generic, not-preprocessing specific tool.


examples/pipeline/plot_multiview_construction.py
```diff
@@ -0,0 +1,96 @@
...
+rgp_scores = []
+clf_scores = []
+
+# The data has 4096 features. The following subspace dimensions are used
+dims = [16, 64, 256, 1024]
+
+skf = StratifiedKFold(n_splits=5, random_state=0, shuffle=True)
+skf.get_n_splits(X, y)
+
+for i, (test_index, train_index) in enumerate(skf.split(X, y)):
```
pierreablin, line 54: Can't `sklearn.model_selection.cross_val_score` be used here, in order to avoid this outer loop?

mvlearn/compose/wrap.py
```diff
@@ -0,0 +1,157 @@
+"""Singleview function wrapping utilities."""
```
pierreablin, line 1: this wrapper is very similar to `mvlearn.preprocessing.ViewTransformer`, so maybe both should be put in the same file?

mvlearn/compose/wrap.py
```diff
@@ -0,0 +1,157 @@
...
+
+        sample_weight : array-like of shape (n_samples,), default=None
+            Sample weights.
+
+        Returns
+        -------
+        score : float
+            Mean accuracy of self.predict(Xs) w.r.t. y
+        """
+        from sklearn.metrics import accuracy_score
```
pierreablin, line 156: why import here?

mvlearn/compose/wrap.py
rflperry: Yes, that was my thought. Here, I am providing a wrapper for estimators with the `predict` function which doesn't belong in a preprocessing module. I put it in `compose` as it's composing a multiview function from sklearn single view ones. I propose moving `ViewTransformer` to `compose`. thoughts?

mvlearn/compose/wrap.py
rflperry: I was following what the `sklearn.ClassifierMixin` class does, but I don't have an explanation for that or see a need to do nested, will change.

mvlearn/compose/wrap.py
pierreablin: Yes I think it's the most sensible thing to do.

mvlearn/compose/wrap.py
```diff
@@ -0,0 +1,278 @@
...
+
+    Examples
+    --------
+    >>> from mvlearn.datasets import load_UCImultifeature
+    >>> from mvlearn.compose import ViewClassifier
+    >>> from sklearn.linear_model import LogisticRegression
+    >>> Xs, y = load_UCImultifeature()
+    >>> clfs = ViewClassifier(LogisticRegression())
+    >>> y_hat = clfs.fit(Xs, y).predict(Xs)
+    >>> print(y.shape)
```
gavinmischler, line 112: Do you want to print y_hat.shape?

mvlearn/compose/wrap.py
```diff
@@ -0,0 +1,278 @@
...
+            Mean accuracy of self.predict(Xs) w.r.t. y
+        """
+        return accuracy_score(y, self.predict(Xs), sample_weight=sample_weight)
+
+
+class ViewTransformer(BaseWrapper, TransformerMixin):
+    r"""Apply a sklearn transformer to each view of a dataset
+
+    Build a transformer from multiview dataset to multiview dataset by
+    using individual scikit-learn transformer on each view.
```
gavinmischler, line 169: using "an" individual...

mvlearn/compose/wrap.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        Return the predicted class labels using majority vote of the
+        predictions from each view.
+
+        Parameters
+        ----------
+        Xs : list of array-likes or numpy.ndarray
+             - Xs length: n_views
+             - Xs[i] shape: (n_samples, n_features_i)
+            The data to predict
```
gavinmischler, line 125: There are lots of places in these functions where Xs bullet points have indentation such that they don't render very well on the docs.

https://190-206838300-gh.circle-artifacts.com/0/dev/references/compose.html#viewclassifier

This is probably how we should try to do it from now on: (taken from SimpleSplitter, which renders well)

        Xs : list of numpy.ndarray
            - Xs length: n_views
            The input multiview dataset

Which results in clean bullet points for its inverse_transform() https://190-206838300-gh.circle-artifacts.com/0/dev/references/compose.html#simplesplitter

This is probably something we can revisit for the documentation as a whole later, because there are definitely lots of places where the change from the old rendering to the new one has messed up the cleanliness of the formatting, but for now let's just change the ones on this PR at least.

gavinmischler (approved):
Just some docs stuff but looks fine

PR #268, opened 2020-12-02
0.4.0 changes

rflperry:
Adds authors, and updates version number and setup.py in preparation for PyPi update.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -101,7 +101,6 @@ Python package dependencies
 - graspy >=0.1.1
 - matplotlib >=3.0.0
 - numpy >=1.17.0
-- pandas >=0.25.0
 - scikit-learn >=0.19.1
 - scipy >=1.1.0
 - seaborn >=0.9.0
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -105,7 +105,6 @@ Python package dependencies
 - scipy >=1.1.0
 - seaborn >=0.9.0
 - joblib >=0.11
-- python-picard >= 0.4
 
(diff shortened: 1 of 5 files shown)


docs/install.rst
```diff
@@ -101,7 +101,6 @@ Python package dependencies
 - graspy >=0.1.1
 - matplotlib >=3.0.0
 - numpy >=1.17.0
-- pandas >=0.25.0
 - scikit-learn >=0.19.1
 - scipy >=1.1.0
 - seaborn >=0.9.0
```
gavinmischler, line 106: It wouldn't let me comment on the python-picard line so I'm doing it here. python-picard is currently in the requirements/multiviewica.txt requirements file, but it's listed here as a core mvlearn dependency, as well as part of the optional [multiviewica]. We should pick one or the other. If it's desirable to not add another dependency, then let's make it part of the optional [multiviewica] dependencies, though we could include it for conda since it's on conda

pierreablin: I think that picard is the default solver for the group ICA functions as well. I am the maintainer of this package, it is both very light and will never change api, so I think its ok to keep it as a core package.

gavinmischler: Ok, if it's going to be a core dependency then it should be moved from requirements/multiviewica.txt to requirements/base.txt

rflperry: good catch, it is incorrectly there in the `.rst` file but correctly only in the multiviewica requirements file.

gavinmischler: wait so are we not putting it as a core dependency?

Issue #278, opened 2021-01-18
CCA stats error if more components than dimensions


rflperry:
Bug provided by John Conroy.

## Error
CCA can compute more canonical components than the minimum dimension of the 2 views (`n_components=None` will do this). I don't think this is an issue.

In `embed.cca.stats`,  only up to the minimum dimension across views is used to compute Rao's F statistic which gives a dimension mismatch error.

## Template Code
```mvlearn
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""
Created on Thu Jan 14 16:19:29 2021

@author: jconroy
"""

# License: MIT
from mvlearn.datasets import sample_joint_factor_model
from mvlearn.embed import CCA, MCCA, KMCCA
from mvlearn.plotting import crossviews_plot


def fit_mvlearn_model(Xs, model_type='CCA', n_components=None):
    '''


    Parameters
    ----------
    Xs : ndarray
        DESCRIPTION.
    model_type : TYPE, optional
        DESCRIPTION. The default is 'CCA'.

    Returns
    -------
    cca
    Xs transformed.

    '''

    ###############################################################################
    # CCA
    # ^^^
    #
    # CCA, equivalent to 2 view MCCA, learns transformations of the views,
    # projecting a linear combination of the features to a component such that the
    # sum of correlations between the ith components of each view is maximized. We
    # see the top three components of the first two views plotted against each
    # other, pairwise. The strong linear shape on the diagonals shows that the
    # found components correlate well.

    # the default is no regularization meaning this is SUMCORR-AVGVAR MCCA
    if model_type == 'CCA':
        cca = CCA(n_components=n_components)
        ###############################################################################
    # Regularized CCA
    # ^^^^^^^^^^^^^^^^
    #
    # We can add regularization with the `regs` argument to handle
    # high-dimensional data. This data is simple, and so it makes little
    # difference. Here, we use MCCA for all 3 views.

    # regularization value of .5 for each view
    elif model_type == 'MCCA':
        cca = MCCA(n_components=joint_rank)  # , signal_ranks=[2, 2, 2])
    ###############################################################################
    # Informative MCCA: PCA then MCCA
    # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    #
    # We can also handle high-dimensional data with i-MCCA. We first compute a
    # low rank PCA for each view, then run MCCA on the reduced data. With rank 2
    # PCA, only the first two CCA components are informative, as we can see in the
    # plot.

[...]

PR #279, opened 2021-01-19
Fixes CCA to only take integer n_components values

rflperry:
#### Reference Issues/PRs
Addresses #278 which was the result of an error with CCA

Previously CCA inherited all `n_components` arguments from MCCA but for two views only integers from 1 to the minimum number of features among views is permissible. >2 views this is not the case.

### DIFF
--- a/mvlearn/embed/cca.py
+++ b/mvlearn/embed/cca.py
@@ -18,13 +18,11 @@ class CCA(MCCA):
 
     Parameters
     ----------
-    n_components : int | 'min' | 'max' | None (default 1)
-        Number of final components to compute. If `int`, will compute that
-        many. If None, will compute as many as possible. 'min' and 'max' will
-        respectively use the minimum/maximum number of features among views.
+    n_components : int (default 1)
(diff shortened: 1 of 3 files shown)


mvlearn/embed/cca.py
```diff
@@ -109,6 +107,11 @@ def _fit(self, Xs):
             raise ValueError(
                 f"CCA accepts exactly 2 views but {self.n_views_}"
                 "were provided. Consider using MCCA for more than 2 views")
+        if not (type(self.n_components) == int and
```
agramfort, line 110: `type(..) == ` is evil. You should use isinstance and for testing if it's an int you should use numbers.Integral https://docs.python.org/fr/3/library/numbers.html#numbers.Integral

```
In [5]: type(np.int64(2)) == int
Out[5]: False

In [6]: import numbers

In [7]: isinstance(np.int64(2), numbers.Integral)
Out[7]: True
```

agramfort (approved):

PR #290, opened 2021-03-23
Add support for memmaps in check arrays

### LINKED ISSUE #289: check_arrays could support memmaps
When multiple datasets are used, memory can be overloaded easily unless one is careful about the implementation.
I think it makes sense to assume that:
- Reduced data can hold in memory
- One dataset can hold in memory
I would like to modify the check_arrays function so that it supports memory maps. I may also change some code in GroupPCA so that it uses reduced amount of memory.
Do you think it is a good idea ?

hugorichard:
Fixes #289

### DIFF
--- a/mvlearn/utils/utils.py
+++ b/mvlearn/utils/utils.py
@@ -82,7 +82,11 @@ def check_Xs(
             )
             raise ValueError(msg)
 
-    Xs = [check_array(X, allow_nd=False, copy=copy) for X in Xs]
+    if type(Xs[0]) != np.memmap:
+        Xs = [check_array(X, allow_nd=False, copy=copy) for X in Xs]
+    else:
+        for X in Xs:
+            check_array(X)
 
     if not len(set([X.shape[0] for X in Xs])) == 1:
         msg = "All views must have the same number of samples"


mvlearn/utils/utils.py
```diff
@@ -82,7 +82,11 @@ def check_Xs(
             )
             raise ValueError(msg)
 
-    Xs = [check_array(X, allow_nd=False, copy=copy) for X in Xs]
+    if type(Xs[0]) != np.memmap:
+        Xs = [check_array(X, allow_nd=False, copy=copy) for X in Xs]
+    else:
+        for X in Xs:
+            check_array(X)
```
agramfort, line 89: not clear what you want to do here. Avoid a copy when memmap? why ignoring the copy param for memmap I find this behavior hard to document. Can we update sklearn check_array to allow memmap and enforce here that copy=False when it's a memmap?

hugorichard: The problem I'd like to avoid is that when you do check_array with a memmap, it returns an array (and not a memmap even when copy=False).
But you are right it would be better to do a PR in sklearn so that it supports memmap. I'll close this then.

agramfort:
+1

we can do this hack here until it gets release in sklearn

>

PR #291, opened 2021-03-24
Add indexes to groupica and grouppca

hugorichard:
Sometimes, you fit a model on many datasets but you want to use transform only on a subset of datasets.
This pull request makes this possible for GroupICA and GroupPCA.
It does so by adding a parameter indexes in the transform method so that you can specify which datasets are to be transformed.


mvlearn/decomposition/groupica.py
```diff
@@ -12,12 +12,10 @@
...
-    msg = (
-        f"ModuleNotFoundError: {error}. multiviewica dependencies"
-        + "required for this function. Please consult the mvlearn"
-        + "installation instructions at https://github.com/mvlearn/mvlearn"
-        + "to correctly install multiviewica dependencies."
-    )
+    msg = (f"ModuleNotFoundError: {error}. multiviewica dependencies" +
+           "required for this function. Please consult the mvlearn" +
+           "installation instructions at https://github.com/mvlearn/mvlearn" +
+           "to correctly install multiviewica dependencies.")
```
agramfort:
```suggestion
    msg = (f"ModuleNotFoundError: {error}. multiviewica dependencies "
           "required for this function. Please consult the mvlearn "
           "installation instructions at https://github.com/mvlearn/mvlearn "
           "to correctly install multiviewica dependencies.")
```

mvlearn/decomposition/groupica.py
```diff
@@ -224,6 +224,12 @@ def transform(self, Xs, y=None):
         y : None
             Ignored variable.
 
+        indexes: None, or np array
+            This has to be used when less views are used in
+            transform than during the fit.
+            View Xs[i] should correspond to the view indexes[i]
+            in the fit
```
agramfort, line 231: I don't like this. but I understand the need. In this case would be accept something like

`Xs = [X[0], None, X[1]]`

so you pass a list and some views can be None?

rflperry: This had been included in the `BaseCCA` implementation, see https://github.com/mvlearn/mvlearn/blob/688548b55ca3b61ea4bcc3b339cee140d03c0adc/mvlearn/embed/base.py#L139

allowing for a single view to be transformed.

hugorichard: Ok so @rflperry  you think the current approach is the one to keep ?
@agramfort we could support the None in the list if you prefer !

rflperry: Not really a downside to supporting None that I can see. And adds valuable utility.

hugorichard: Ok so let's do that instead. I'll update the API.

hugorichard: I would rather have both things supported actually. It is much simpler to specify indexes than having to had None everywhere.

hugorichard: @agramfort There is also the behavior in inverse_transform that should be discussed. I think indexes are good because it only returns arrays that are wanted and in this case it is not obvious to use the None trick

rflperry: I am content with `transform` having an `indexes` (perhaps `indices`) parameter which allows either an integer or a list of integers. I can see potential scenarios where allowing `None` views in `Xs` but I don't think it's a necessary implementation.

hugorichard: So if @agramfort agrees, I'd like to keep the `indexes` parameter so that `transform` and `inverse_transform` use the same API and delay the support of None to another pull request.
I'll change its name to indices

agramfort:
what worries me is that if we do this need to be consistent in terms of API.

The api for all transformers becomes

transform(self, Xs, indexes=None)
fit_transform(self, Xs, y, indexes=None)

and then it needs to be dealt with in the pipeline mechanism etc.
Everytime we added parameters in fit, transform in sklearn it created
issues and inconsistencies in the long term.

>

rflperry:
If mvlearn were to add a single parameter to `fit` and `fit_transform`, I think an `indices` parameter would be it and that basically bridges the gap between single and multiview algorithms. That being said, there may be consequences to such a parameter that I'm not thinking of but I think it should be compatible with the way things work now.

A PR would have to update the API and update every current function, or at least be part of a transition to that new API. The alternatives are allowing inputs of `None` in `Xs` but that seems like a pain for the user to add as input and then extract an output. Maybe a decent temporary short term fix.

hugorichard:
Fit and fit transform would remain the same. I think only transform and inverse transform need indexes.

agramfort:
you're in charge.

>

rflperry:
@hugorichard I say go ahead and make the indices changes in this PR. Update the API with the disclaimer that it's not implemented everywhere yet. I will help with making the change in other methods.

hugorichard:
@rflperry I made the change of API and changed the name of indexes to index.
I don't get why make html throws an error. It seems completely unrelated to the changes in this PR...

mvlearn/decomposition/groupica.py
```diff
@@ -235,19 +242,29 @@ def transform(self, Xs, y=None):
         """
         check_is_fitted(self)
         Xs = check_Xs(Xs, copy=True)
+        if index is None:
+            index_ = np.arange(len(self.means_))
```
rflperry, line 246:
```suggestion
            index_ = np.arange(self.n_views_)
```
Might as well use `n_views_` if we store it

mvlearn/decomposition/grouppca.py
```diff
@@ -262,15 +269,42 @@ def transform(self, Xs, y=None):
         """
         check_is_fitted(self)
         Xs = check_Xs(Xs, copy=True)
+
+        if index is None:
+            index_ = np.arange(len(self.individual_projections_))
```
rflperry, line 274:
```suggestion
            index_ = np.arange(self.n_views_)
```

mvlearn/decomposition/grouppca.py
```diff
@@ -304,24 +338,48 @@ def inverse_transform(self, X_transformed):
...
+            mvlearn yet.
+
         Returns
         -------
         Xs : list of arrays
             The recovered individual datasets
         """
         check_is_fitted(self)
+        if index is None:
+            index_ = np.arange(len(self.individual_projections_))
```
hugorichard, line 355:
```suggestion
            index_ = np.arange(len(self.n_views_))
```

mvlearn/decomposition/grouppca.py
```diff
@@ -262,15 +269,42 @@ def transform(self, Xs, y=None):
...
-                        self.individual_projections_, Xs, self.individual_mean_
+                        [self.individual_projections_[i] for i in index_],
+                        Xs,
+                        [self.individual_mean_[i] for i in index_],
                     )
                 )
             ]
+
+        if index is not None:
+            return np.mean(
```
rflperry, line 291: For legibility, since the difference between `self.multiview_output` being `True` vs `False` is just the mean, I would compute the list of  `[np.dot(X - mean, W.T) for ....]` outside of the `if self.multiview_output:` and then either return it or return its mean.

mvlearn/decomposition/grouppca.py
```diff
@@ -304,24 +338,48 @@ def inverse_transform(self, X_transformed):
...
                     zip(
                         X_transformed,
-                        self.individual_embeddings_,
-                        self.individual_mean_,
+                        [self.individual_embeddings_[i] for i in index_],
+                        [self.individual_mean_[i] for i in index_],
+                    )
+                )
+            ]
+        if index is not None:
```
rflperry, line 372: Same comment as above in terms of repeated code.

rflperry:
I don't see an html error. Do you recall what was failing?

mvlearn/decomposition/grouppca.py
hugorichard: The situation is a bit different here as we do not iterate over X_transformed

mvlearn/decomposition/groupica.py
```diff
@@ -224,6 +224,13 @@ def transform(self, Xs, y=None):
         y : None
             Ignored variable.
 
+        index: None, or int or array-like
+            int or list of ints specifying the indices of the
+            inputted views relative to the fitted views.
+            If None, there should be as many inputted views as fitted views.
+            Note that the index parameter is not available in all methods of
+            mvlearn yet.
```
rflperry, line 232:
```suggestion
        index: int or array-like, default=None
            The index or list of indices of the fitted views to which the
            inputted views correspond. If None, there should be as many
            inputted views as the fitted views and in the same order.
            Note that the index parameter is not available in all methods of
            mvlearn yet.
```

Just want to get the language right here, as it will be what I use for everything. If looks good to you, add to everywhere else in this PR please.

rflperry:
LGTM otherwise thanks! I'll make these changes elsewhere so we are consistent.

rflperry (approved):

hugorichard:
Thanks @rflperry ! I'll do the same changes in multiviewICA and ViewTransformer. I have a PR almost ready for these.

hugorichard:
> I don't see an html error. Do you recall what was failing?

It may be only me but it looks like the CI build fails (build_docs). It throws the error:
`

Extension error:
Here is a summary of the problems encountered when running the examples

Unexpected failing examples:
/home/circleci/project/examples/embed/plot_omnibus_embedding.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/circleci/.local/lib/python3.7/site-packages/sphinx_gallery/scrapers.py", line 333, in save_figures
    rst = scraper(block, block_vars, gallery_conf)
  File "/home/circleci/.local/lib/python3.7/site-packages/sphinx_gallery/scrapers.py", line 149, in matplotlib_scraper
    fig.savefig(image_path, **these_kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/figure.py", line 2959, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/backend_bases.py", line 2261, in print_figure
    **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/backend_bases.py", line 1669, in wrapper
    return func(*args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/backends/backend_agg.py", line 508, in print_png
    FigureCanvasAgg.draw(self)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/backends/backend_agg.py", line 406, in draw
    self.figure.draw(self.renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/artist.py", line 74, in draw_wrapper
    result = draw(artist, renderer, *args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/artist.py", line 51, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/figure.py", line 2735, in draw
    renderer, self, artists, self.suppressComposite)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/image.py", line 132, in _draw_list_compositing_images
    a.draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/artist.py", line 51, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/_api/deprecation.py", line 431, in wrapper
    return func(*inner_args, **inner_kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 2925, in draw
    mimage._draw_list_compositing_images(renderer, self, artists)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/image.py", line 132, in _draw_list_compositing_images
    a.draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/artist.py", line 51, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/legend.py", line 614, in draw
    self._legend_box.draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/offsetbox.py", line 368, in draw
    c.draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/offsetbox.py", line 368, in draw
    c.draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/offsetbox.py", line 368, in draw
    c.draw(renderer)
  [Previous line repeated 1 more time]
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/offsetbox.py", line 694, in draw
    c.draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/artist.py", line 51, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/collections.py", line 1009, in draw
    super().draw(renderer)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/artist.py", line 51, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/collections.py", line 380, in draw
    len(facecolors) == 1 and len(edgecolors) == 1 and
TypeError: object of type 'NoneType' has no len()


/home/circleci/project/examples/embed/plot_cca_comparison.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/circleci/project/examples/embed/plot_cca_comparison.py", line 104, in <module>
    ax.set_xticks([], [])
  File "/home/circleci/.local/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 73, in wrapper
    return get_method(self)(*args, **kwargs)
TypeError: set_ticks() takes 2 positional arguments but 3 were given


-------------------------------------------------------------------------------
Makefile:22: recipe for target 'html' failed
make: *** [html] Error 2

Exited with code exit status 2

CircleCI received exit code 2
`

rflperry:
I didn't see this error when it had rendered before I made comments but I see it now. Let me look into it, I think there was a potential deprecation for `ax.set_ticks([], [])` but I don't think it is deprecated. The omnibus embedding issue I'm unsure of.

rflperry:
matplotlib says their latest stable version is `3.3.4` but the pypi version is `3.4.0` which is what we are installing. They may have just done an update which is causing problems.

rflperry:
@hugorichard I confirmed that I experienced this bug locally with matplotlib version `3.4.0` but it runs fine with version `3.3.4`. Please update the `requirements/base.txt` file accordingly and I will make an issue.

rflperry:
You can merge when you are ready.

hugorichard:
I don't have write access.

hugorichard:
Thanks

PR #297, opened 2021-09-27
CV: GridSearchCV and RandomizedSearchCV

jameschapman19:
Adds some classes that wrap the scikit-learn BaseSearchCV classes and allow hyperparameter tuning in the same style using gridsearch and randomized search with cv

#### Any other comments?
Basically the barrier to using GridSearchCV and RandomizedSearchCV out of the box is that we sometimes have different values for the same parameter for different views as well as the similar problem to cross_validate() where we need to put in a splitter.

What I have done is to edit the fit() method in the scikit BaseSearchCV class and then made GridSearchCV and RandomizedSearchCV inherit the edited class with its new fit method.

On the parameter inputs side I have added a small utility function param2grid which essentially converts parameter grids containing multiple values for each view into a scikit friendly form. I think this cleans up the input from a user point of view as they can do something like:

`grid={'parameter': [[view 1 grid][view 2 grid]]}`

as opposed to

`grid={'parameter' : [list of all combinations of view 1 and view 2 grids]}`

I have also rejigged the scikit parameter sampler slightly to solve similar problems.

#### Noting Urgently!
I would like to add in the appropriate references to original code as large parts are exactly the same as scikit-learn. Didn't expect this pull request to go to the main repo expected it just to my fork!

### DIFF
--- a/mvlearn/model_selection/__init__.py
+++ b/mvlearn/model_selection/__init__.py
@@ -1,4 +1,5 @@
 from .split import train_test_split
 from .validation import cross_validate
+from ._search import GridSearchCV, RandomizedSearchCV
 
-__all__ = ["train_test_split", "cross_validate", ]
+__all__ = ["train_test_split", "cross_validate", "GridSearchCV", "RandomizedSearchCV"]
--- a/mvlearn/model_selection/_search.py
+++ b/mvlearn/model_selection/_search.py
@@ -117,11 +117,11 @@ class ParameterSampler:
             for key in dist:
                 if isinstance(dist[key], Iterable):
                     if any(
-                        [
-                            not isinstance(view_param, Iterable)
-                            and not hasattr(view_param, "rvs")
-                            for view_param in dist[key]
-                        ]
+                            [
+                                not isinstance(view_param, Iterable)
+                                and not hasattr(view_param, "rvs")
+                                for view_param in dist[key]
+                            ]
                     ):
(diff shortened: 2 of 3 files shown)


rflperry:
Is it possible to just use [itertools.product](https://docs.python.org/3/library/itertools.html#itertools.product) and pass in the mvlearn CV splitter to the sklearn Gridsearch?

jameschapman19:
Yep you're right!

```
def main():
    from sklearn.pipeline import Pipeline
    from mvlearn.compose import SimpleSplitter
    from mvlearn.embed import MCCA
    n_samples = 100
    n_features = [2, 3, 4]
    rng = np.random.RandomState(0)
    Xs = [rng.randn(n_samples, n_feature) for n_feature in n_features]
    c1 = [0.1, 0.2]
    c2 = [0.2, 0.3]
    c3 = [0.3, 0.4]

    param_grid = {
        'estimator__regs': list(map(list, itertools.product(c1, c2, c3))),
    }
    estimator = MCCA()
    pipeline = Pipeline([('splitter', SimpleSplitter(n_features)),
                         ('estimator', estimator)])
    def scorer(estimator, X):
        scores = estimator.score(X)
        return np.mean(scores)

    cv_model = GridSearchCV(pipeline, cv=3, param_grid=param_grid, scoring=scorer).fit(np.hstack(Xs))

if __name__ == '__main__':
    main()
```

Issue #307, opened 2022-03-28
CircleCI Pipeline is broken


gavinmischler:
## Expected Behavior
CircleCI should be able to build the package and documentation on a PR.

## Actual Behavior
Currently, CircleCI is broken due to the recent changes in Github's authentication requirements. It is no longer able to merge with upstream without an authentication token. This has caused the build to fail on recent PRs #306 and #304 

## How to Fix
I believe we will need to create a key for the repo that CircleCI can use to access mvlearn. This key should be stored on CircleCI, not in the repo anywhere public. [This tutorial](https://blog.lelonek.me/private-dependencies-from-github-in-your-docker-container-92e3b8cbf677) might be helpful. However, I don't think I have permission on mvlearn to create a token, and I'm not sure if I can change the settings on CircleCI either. @rflperry are you able to do this?

## Full Traceback from CircleCI
```pytb
#!/bin/bash -eo pipefail
echo $(git log -1 --pretty=%B) | tee gitlog.txt
echo ${CI_PULL_REQUEST//*pull\//} | tee merge.txt
if [[ $(cat merge.txt) != "" ]]; then
  echo "Merging $(cat merge.txt)";
  git remote add upstream git://github.com/mvlearn/mvlearn.git;
  git pull --ff-only upstream "refs/pull/$(cat merge.txt)/merge";
  git fetch upstream main;
fi

remove graspy, omnibus, and tests
306
Merging 306
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.

Exited with code exit status 1
CircleCI received exit code 1

```

rflperry:
Hmm okay yes it was previously failing at the [installation phase](https://app.circleci.com/pipelines/github/mvlearn/mvlearn/317/workflows/c48d204c-015f-4c92-8ada-1de4c1ce2b34/jobs/377) and I wasn't sure why it started failing but it seemed to be due to running out of memory. I will take a look at this separate issue, but the previous one still remains.

PR #308, opened 2022-03-29
Fix CircleCI pipline: issue #307

rflperry:
Fixes Issue #307 by switching git:// to https://. the former is no longer supported ([see here](https://github.blog/2021-09-01-improving-git-protocol-security-github/)).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
               echo ${CI_PULL_REQUEST//*pull\//} | tee merge.txt
               if [[ $(cat merge.txt) != "" ]]; then
                 echo "Merging $(cat merge.txt)";
-                git remote add upstream git://github.com/mvlearn/mvlearn.git;
+                git remote add upstream https://github.com/mvlearn/mvlearn.git;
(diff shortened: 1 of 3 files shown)


rflperry:
@gavinmischler These changes fix issue #307 but as you can see it fails to build the docs due to exceeding the maximum time limit (10 min) which is an issue. Notably though the previous issue of failing to install the python requirements is no longer an issue.

I recommend merging these changes which should allow for tests on the other PRs.

Tutorials for `plot_mv_spectral_validation_simulated.py` and `plot_mv_coregularized_spectral_tutorial.py` I believe were sources of time issue but I'm not fully sure.

gavinmischler:
I see that the CI tests for py38 failed because of a couple tests in test_mcca, and I actually had the same issue for one of my other recent PRs. I just re-pushed to the branch to make it run the tests again and they passes, but it seems there is some randomness introduced in that test, which isn't ideal. Not sure if it's a coincidence, but mine also only failed py38, so maybe there's something in py38 that changed some behavior to be random

gavinmischler (approved):

gavinmischler:
@rflperry I'm ok with merging this still so we can diagnose the other issues. Are you able to override the failed checks and merge?

PR #311, opened 2022-03-30
Increase build docs timeout to 30m

### LINKED ISSUE #310: CircleCI and docs fails since over the time limit (10 min)
`Too long with no output (exceeded 10m0s): context deadline exceeded`

I believe an issue with run time of cluster scripts.

gavinmischler:
#### Reference Issues/PRs

Fixes #310 

Increases timeout limit from 10m (default) to 30m since the make html command is failing simply because it takes too long to build all the docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
               echo ${CI_PULL_REQUEST//*pull\//} | tee merge.txt
               if [[ $(cat merge.txt) != "" ]]; then
                 echo "Merging $(cat merge.txt)";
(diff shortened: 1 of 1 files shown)


gavinmischler:
@rflperry it looks like the issue was just that the docs took too long to build overall, not that a specific example script was hanging or timing out. This should fix that

rflperry:
Yep, I just didn't realize you could increase the time limit. It takes ~14 min which is probably too long though in the longer term. But I'm happy to make the change so things pass for now.

PR #314, opened 2022-04-01
Explicit requirements for conda version

gavinmischler:
This is needed for the conda-forge version

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,9 +30,6 @@ REQUIRED_PACKAGES = [
     "scipy>=1.5.0",
     "seaborn>=0.9.0",
     "joblib>=0.11",
-    "torch>=1.1.0",
-    "tqdm",
-    "python-picard >= 0.4",
 ]
 
 # Find mvlearn version.
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ AUTHOR_EMAIL = "rflperry@gmail.com"
 URL = "https://github.com/mvlearn/mvlearn"
 MINIMUM_PYTHON_VERSION = 3, 6  # Minimum of Python 3.6
 REQUIRED_PACKAGES = [
-    "matplotlib>=3.0.0, <=3.3.4",
+    "matplotlib>=3.0.0",
     "numpy>=1.17.0",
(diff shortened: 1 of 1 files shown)


rflperry (approved):

agramfort (commented):
Installing torch via pip is dangerous. It often leads to broken gpu installs

rflperry:
Since we are leaving out `multiviewica_extras` anyways, we could leave out the torch extas too. I kind of prefer that since it's not a mandatory package to install.

In the conda forge installation, the required package dependencies are installted via pip or conda?

gavinmischler:
I believe all dependencies are installed through conda-forge. That’s why we currently leave out multiviewica, since it doesn’t have a conda-forge installation. So this will install torch through conda-forge

gavinmischler:
But we could totally remove all the extra requirements and make the conda version more in line with the pip version of mvlearn, and then just provide instructions for installing the correct dependencies through conda if a user wants to use them

rflperry:
yeah I would say it's probably better to not necessitate torch as a dependency. that's a bit bulky.

agramfort:
sounds good !

setup.py
```diff
@@ -23,16 +23,14 @@
...
-with open("./requirements/torch.txt", "r") as f:
-    torch_extras = f.read()
-with open("./requirements/multiviewica.txt", "r") as f:
-    multiviewica_extras = f.read()
-EXTRA_PACKAGES = {
-    'torch': torch_extras,
-    'multiviewica': multiviewica_extras,
-}
+REQUIRED_PACKAGES = [
+    "matplotlib>=3.0.0, <=3.3.4",
```
agramfort, line 27: can we avoid the <=3.3.4? I don't like pinning versions in libraries or limiting max version and it leads to env that cannot be updated

gavinmischler: I believe this was changed because some of the examples notebooks broke for 3.4 and above. There is nothing in mvlearn itself that requires the maximum, so it should be fine to remove the <=3.3.4 from here and then work on updating the example notebooks as a follow-up

PR #318, opened 2023-10-20
CTClassifier fix

### LINKED ISSUE #317: AttributeError: 'CTClassifier' object has no attribute 'estimator1'
Hi everyone,

the CTClassifier failed to run as no attribute 'estimator1' was found. Can anyone fix this issue?

gavinmischler:
#### Reference Issues/PRs

Fixes #317 

Changes the attribute names of estimator1 and estimator2 in CTClassifier and CTRegressor to not have trailing underscores, which allows them to be properly picked up by sklearn's get_params() method. Also editted other init args which were set as attributes with trailing underscores.

### DIFF
--- a/mvlearn/semi_supervised/ctclassifier.py
+++ b/mvlearn/semi_supervised/ctclassifier.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 from sklearn.naive_bayes import GaussianNB
+from copy import deepcopy
 
 from .base import BaseCoTrainEstimator
 from ..utils.utils import check_Xs, check_Xs_y_nan_allowed
@@ -34,7 +35,7 @@ class CTClassifier(BaseCoTrainEstimator):
         Does not need to be of the same type as ``estimator1``, but should
         support predict_proba().
 
-    p : int, optional (default=None)
+    positive_samples : int, optional (default=None)
         The number of positive classifications from the unlabeled_pool
         training set which will be given a positive "label". If None, the
         default is the floor of the ratio of positive to negative examples
@@ -43,7 +44,7 @@ class CTClassifier(BaseCoTrainEstimator):
         labels are 0 or 1, positive is defined as 1, and in general, positive
         is the larger label.
 
-    n : int, optional (default=None)
+    negative_samples : int, optional (default=None)
         The number of negative classifications from the unlabeled_pool
         training set which will be given a negative "label". If None, the
         default is the floor of the ratio of positive to negative examples
@@ -67,19 +68,19 @@ class CTClassifier(BaseCoTrainEstimator):
     Attributes
     ----------
     estimator1 : classifier object
-        The classifier used on view 1.
+        The classifier to be used on view 1.
 
     estimator2 : classifier object
-        The classifier used on view 2.
+        The classifier to be used on view 2.
 
     class_name_: string
         The name of the class.
 
-    p : int, optional (default=None)
+    positive_samples : int, optional (default=None)
         The number of positive classifications from the unlabeled_pool
         training set which will be given a positive "label" each round.
 
-    n : int, optional (default=None)
+    negative_samples : int, optional (default=None)
         The number of negative classifications from the unlabeled_pool
         training set which will be given a negative "label" each round.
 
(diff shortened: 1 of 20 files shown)


mvlearn/semi_supervised/ctclassifier.py
```diff
@@ -178,20 +178,20 @@ def __init__(self,
...
             n = p
-            self.p_, self.n_ = p, n
+            self.p, self.n = p, n
         elif (p is None and n is not None):
             p = n
-            self.p_, self.n_ = p, n
+            self.p, self.n = p, n
         else:
-            self.p_, self.n_ = p, n
+            self.p, self.n = p, n
```
agramfort: this should be done in the fit. There should be no logic in the init just do:

self.a = a
self.b = b

etc.

mvlearn/semi_supervised/ctclassifier.py
```diff
@@ -305,30 +305,30 @@ def fit(self, Xs, y):
                 it += 1
 
                 # fit each model to its respective view
-                self.estimator1_.fit(X1[L], y[L])
-                self.estimator2_.fit(X2[L], y[L])
+                self.estimator1.fit(X1[L], y[L])
+                self.estimator2.fit(X2[L], y[L])
```
agramfort: the fit should not change the init parameters but just add / modify attributes that end with _. So you should create 2 new attributes here called estimator1_ and estimator2_ (on top of estimator1 and estimator2)

mvlearn/semi_supervised/ctclassifier.py
```diff
@@ -271,14 +271,14 @@ def fit(self, Xs, y):
         if self.n_classes > 1:
 
             # if both p & n are none, set as ratio of one class to the other
-            if (self.p_ is None and self.n_ is None):
```
agramfort, line 274: also I would avoid single letter attributes. Use n_samples_ and n_features_ ?

gavinmischler: Changed them to ``positive_samples`` and ``negative_samples``

gavinmischler:
@agramfort I believe I've addressed your comments. Regarding the tests failing, I opened another PR because the tests are from other dependencies having things deprecated, unrelated to CTClassifier. If we merge that one first it will fix the tests here.

Also, regarding the CircleCI docs build failing, it's failing because CircleCI deprecated the python docker image we were using, but when I tried updating to a new image (``cimg/python:3.18.15``) it still doesn't work. Do you have experience with this?

mvlearn/semi_supervised/ctregression.py
```diff
@@ -278,8 +276,8 @@ def fit(self, Xs, y):
         L2 = L.copy()
 
         # fitting the estimator object on the train data
-        self.estimator1_.fit(X1[L1], y1[L1])
-        self.estimator2_.fit(X2[L2], y2[L2])
+        self.estimator1.fit(X1[L1], y1[L1])
+        self.estimator2.fit(X2[L2], y2[L2])
```
agramfort: this was correct. See self.estimator1 as a kind of factory that you can copy/clone to get an instance you want to use at fit time. The fit method should only modify parameters that end with _. The self.estimator1 should stay untouched. It means you need both self.estimator1 and self.estimator1_ (the clean one from init you could reuse for a new fit on a new dataset) and the one used at fit one the last data.

gavinmischler: got it, I had made the change in ctclassifier but not ctregression. Now I make a copy of the original estimator1 and estimator2 each time ``fit`` is called

agramfort (approved):
LGTM

no idea how hard it is at this point to make the CIs green

gavinmischler:
I got the docs to build at least and the tests pass, but there seems to be an issue with accessing the built docs artifact. I wonder if this is related to [this PR](https://github.com/mne-tools/mne-python/pull/11622) in MNE which added a CircleCI API token for their CircleCI Artifacts workflow. I don't have the permissions for mvlearn or circleci to create such a token and add it as a secret though, so I don't have a way of testing this.

Also, someone with permissions should remove the PR checklist requirement of the 3.6 tests because this PR removed those in favor of the 3.7 and 3.8 tests but it's still limiting whether the PR can be merged

